PR #58, opened 2016-05-05
Feature/fix pdfs

dustinvtran:
#### Summary:

Issues: makes progress for #8

I fixed and robustified the implementation for Dirichlet, Gamma, Inverse Gamma, and Multinomial distributions. I also let all distributions take np.array or tf.tensors as arguments; previously it was only tf.tensors as arguments with the occasional np argument, which was confusing.

#### Intended Effect:

The above distributions work.

#### How to Verify:

Run all examples and unit tests. See that nothing is broken. Tests are added for the fixed distributions.

#### Side Effects:

N/A

#### Documentation:

Doc is added for any distributions that use vectors as arguments (e.g. dirichlet, multinomial, multivariate normal). Distributions without doc use scalars as arguments by default.

#### Reviewer Suggestions: 

@akucukelbir since you have time? or @dawenl? Someone just has to run the unit tests to verify that it works on their computer too.

### DIFF
--- a/edward/stats/distributions.py
+++ b/edward/stats/distributions.py
@@ -109,6 +109,7 @@ class InvGamma:
                tf.mul(-alpha-1, tf.log(x)) - tf.truediv(beta, x)
 
 class Multinomial:
+    """There is no equivalent version implemented in SciPy."""
     def rvs(self, n, p, size=1):
         return np.random.multinomial(n, p, size=size)
 
@@ -117,18 +118,19 @@ class Multinomial:
         Arguments
         ----------
         x: np.array or tf.Tensor
-            vector
-        mean: np.array or tf.Tensor, optional
-            vector. Defaults to zero mean.
-        cov: np.array or tf.Tensor, optional
-            vector or matrix. Defaults to identity.
+            vector of length K, where x[i] is the number of outcomes
+            in the ith bucket
+        n: int or tf.Tensor
+            number of outcomes equal to sum x[i]
+        p: np.array or tf.Tensor
+            vector of probabilities summing to 1
         """
-        # TODO dimensions?
-        x = tf.squeeze(x)
-        n = tf.squeeze(n)
+        x = tf.cast(tf.squeeze(x), dtype=tf.float32)
+        n = tf.cast(tf.squeeze(n), dtype=tf.float32)
         p = tf.cast(tf.squeeze(p), dtype=tf.float32)
-        return tf.log(factorial(n)) - \
-               tf.log(tf.reduce_prod(factorial(x))) + \
+        one = tf.constant(1.0, dtype=tf.float32)
+        return log_gamma(n + one) - \
+               tf.reduce_sum(log_gamma(x + one)) + \
                tf.reduce_sum(tf.mul(x, tf.log(p)))
 
 class Multivariate_Normal:
@@ -267,7 +269,7 @@ dirichlet = Dirichlet()
 expon = Expon() # TODO unit test
 gamma = Gamma()
 invgamma = InvGamma()
-multinomial = Multinomial() # TODO unit test
+multinomial = Multinomial()
 multivariate_normal = Multivariate_Normal()
 norm = Norm()
 poisson = Poisson() # TODO unit test
--- a/edward/stats/distributions.py
+++ b/edward/stats/distributions.py
@@ -32,6 +32,11 @@ class Distribution:
         -------
         tf.Tensor
             scalar
+
+        Note
+        ----
+        The following distributions use scalar arguments unless
+        documented otherwise.
         """
         raise NotImplementedError()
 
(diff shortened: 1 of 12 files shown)


edward/stats/distributions.py
```diff
@@ -91,18 +105,33 @@ def logpdf(self, x, alpha, beta):
         x = tf.cast(tf.squeeze(x), dtype=tf.float32)
         alpha = tf.cast(tf.squeeze(alpha), dtype=tf.float32)
         beta = tf.cast(tf.squeeze(beta), dtype=tf.float32)
-        return -log_inv_gamma(alpha, beta) - \
-               tf.mul(alpha+1, tf.log(x)) - tf.truediv(beta, x)
+        return tf.mul(alpha, tf.log(beta)) - log_gamma(alpha) + \
+               tf.mul(-alpha-1, tf.log(x)) - tf.truediv(beta, x)
```
akucukelbir: logic/computation change.

edward/stats/distributions.py
```diff
@@ -91,18 +105,33 @@ def logpdf(self, x, alpha, beta):
...
+            vector of probabilities summing to 1
+        """
+        x = tf.cast(tf.squeeze(x), dtype=tf.float32)
+        n = tf.cast(tf.squeeze(n), dtype=tf.float32)
         p = tf.cast(tf.squeeze(p), dtype=tf.float32)
-        return -log_multinomial(x, n) + tf.reduce_sum(tf.mul(x, tf.log(p)))
+        one = tf.constant(1.0, dtype=tf.float32)
+        return log_gamma(n + one) - \
+               tf.reduce_sum(log_gamma(x + one)) + \
+               tf.reduce_sum(tf.mul(x, tf.log(p)))
```
akucukelbir: logic/computation change.

edward/stats/distributions.py
```diff
@@ -164,23 +188,18 @@ def entropy(self, mean=None, cov=1):
...
             det_cov = 1.0
         else:
-            cov = tf.cast(cov, dtype=tf.float32)
+            cov = tf.cast(tf.squeeze(tf.convert_to_tensor(cov)), dtype=tf.float32)
             d = get_dims(cov)[0]
-            if len(cov.get_shape()) == 1: 
+            if len(cov.get_shape()) == 1:
                 det_cov = tf.reduce_prod(cov)
             else:
                 det_cov = tf.matrix_determinant(cov)
```
akucukelbir: eek. it's a shame cholesky is not well supported in tensorflow (yet). this will be numerically unstable for some case in the near future; we should beware.

edward/util.py
```diff
@@ -188,6 +171,9 @@ def logit(x, clip_finite=True):
 
     return transformed, log_jacobian
 
+def multivariate_log_beta(x):
+    return tf.reduce_sum(log_gamma(x)) - log_gamma(tf.reduce_sum(x))
+
```
akucukelbir: do we have tests for these small functions?

tests/test_stats_beta_logpdf.py
```diff
@@ -7,41 +7,38 @@
...
-
-    _test_logpdf_scalar(.3, a=.5, b=5.)
-    _test_logpdf_scalar(.7, a=.5, b=5.)
+    _test_logpdf_scalar(0.3, a=1.0, b=1.0)
+    _test_logpdf_scalar(0.7, a=1.0, b=1.0)
 
-    _test_logpdf_scalar(.3, a=5., b=.5)
-    _test_logpdf_scalar(.7, a=5., b=.5)
+    _test_logpdf_scalar(0.3, a=0.5, b=5.0)
+    _test_logpdf_scalar(0.7, a=0.5, b=5.0)
```
akucukelbir: `a` and `b` reverted here. is that okay?

akucukelbir:
@dustinvtran 

looks great. i added some line comments to bring your attention to "big" changes. as long as they look good to you then i don't expect any further action on them.

i added a small test to the multivariate normal entropy.

the pdf tests pass. (however a bunch of tests in `test_mfgaussian_log_prob_zi.py` fail... we should talk about it.)

edward/stats/distributions.py
dustinvtran: what's a logic/computation change?

edward/util.py
dustinvtran: Nope. Not yet at least.

tests/test_stats_beta_logpdf.py
dustinvtran: Yeah we test both cases in the script, the diff just looks like the arguments are switched.

edward/stats/distributions.py
akucukelbir: as in logic/computation has changed here. just an alert.

edward/util.py
akucukelbir: okay i'll create an issue

Issue #57, opened 2016-05-04
Stochastic Block Model


abojchevski:
Can I define/learn a stochastic block model (SBM) with edward, and if I can what inference engine should I use?

With pymc3 I would define SBM, for some value K (number of blocks/clusters) as such:

```
# group assignment probabilities
pi = Dirichlet(name='pi', a=5 * np.ones(K)) 
# group assignments for nodes
q = Categorical(name='q', p=pi, shape=N) 
# connection probabilities among groups
eta = Beta(name='eta', alpha=1, beta=1, shape=(K, K))
# adjacency matrix
R = Bernoulli(name='R', p=eta[q, :][:, q], observed=R_observed, shape=(N, N))
```

As a probabilistic graphical model it would look like this:
![](https://cloud.githubusercontent.com/assets/8150330/13494649/eb06794a-e145-11e5-9580-c4812645f24d.jpg)

In other words I model my graph (with N nodes and some edges) by modeling the adjacency matrix `R`. Each edge `i, j` is independent from the other edges and is generated as such:
1. sample `pi` from Dirichlet
2. sample `q_i` the categorical cluster assignment for the i-th node given `pi`
3. sample `q_j` the categorical cluster assignment for the j-th node given `pi`
3. from the `K x K` matrix of Beta variables pick the Beta variable corresponding to the cluster assignments (e.g. `q_i=a` and  `q_j=b` => we sample from the Beta variable at index `a, b`)
4. form an edge according to a Bernoulli with probability of 1 given the chosen Beta in the previous step

The adjacency matrix `R` is what is observed, and I would like to learn the latent `eta` and `q`. The MCMC approach with pycm3 works well enough, but a variational approach might be better.

If this kind of question is not suited here please let me know.

dustinvtran:
Hi @abojchevski. A stochastic blockmodel can be fit using the variational inference engine here. To do so here, you would write down the log joint density which is the sum of the log Dirichlet density, Categorical densities, and Bernoulli. We'll have a mixture model example soon to demonstrate this so I'm keeping this issue open.

abojchevski:
@dustinvtran Thanks for the reply. I'll give it a try myself as well.

zfrenchee:
@dustinvtran Hello! I'm also interested in (Mixed Membership) Stochastic Block Models. It seems like the relevant example you post here is outdated (following the link). What might be a relevant example for figuring out how to implement this model in Edward?

dustinvtran:
[...]

Issue #22, opened 2016-02-23
add model examples


dustinvtran:
It'd be great to have high profile model examples which we can highlight on the front page. Some ideas:
- [x] Bayesian linear regression
- [x] Hierarchical generalized linear model
- [x] Mixture model
- [ ] Latent Dirichlet allocation
- [ ] Probabilistic matrix factorization
- [ ] Hidden Markov model
- [ ] Stochastic block model
- [ ] Undirected graphical model/Markov random field
- [ ] Dirichlet process mixture model
- [x] Gaussian process
- [ ] Poisson process
- [x] Bayesian neural network
- [x] Variational auto-encoder
- [x] Deep latent Gaussian model
- [x] Generative adversarial network
- [ ] Bayesian word embedding model
- [ ] Language model with LSTM
- [ ] Sparse-Gamma deep exponential family
- [ ] DRAW

We can think about the choice of inference algorithm and data for the above later.

siddharth-agrawal:
@dustinvtran I want to contribute to Edward and feel that working on this issue will help me more as well. Do you have any suggestions as I go forward? I'm using [this guide](https://github.com/sevntu-checkstyle/sevntu.checkstyle/wiki/Development-workflow-with-Git:-Fork,-Branching,-Commits,-and-Pull-Request) for the dev workflow.

dustinvtran:
perfect! i think a good way to start is maybe to select one of the models you're interested in above and then to implement a working version of it using some data and edward's inference.

siddharth-agrawal:
@dustinvtran I'm thinking of starting with 'Poisson process' next. Do you have any references that might help me? I'm not familiar with many variational inference models.

dustinvtran:
@siddharth-agrawal: Cool. You can try building off https://github.com/blei-lab/edward/pull/294, which I think @diengadji was working on but is no longer(?). It defines a Cox process for spatial data. 

You can look into using the same example or apply a simpler version (e.g., Poisson vs Cox, i.i.d. data instead). I also recommend using a real data set, perhaps one typically applied with Poisson processes.

patricieni:
@dustinvtran After I play a bit more with Edward and look through the code in more depth, I'd like to have a go at implementing a markov random field. Aware of anyone working on undirected models?

dustinvtran:
Undirected is difficult. We haven't focused on it, although there is limited support such as for undirected models that can be conditionally specified (see https://github.com/mariru/exponential_family_embeddings).

[...]

PR #61, opened 2016-05-06
Feature/example mixture gaussian

dustinvtran:
#### Summary:

Issues: makes progress for #7, #22, #57

This adds an example of mixture modeling on synthetic data, using both mean-field variational inference and maximum a posteriori. For mean-field variational inference, the variational family is a product of Dirichlet, Gaussian, and Inverse Gamma densities. 

This also makes progress on how to build an interface allowing mean-field families that are a product of different distributions rather than only one distribution.

#### Intended Effect:

Mixture model examples are available.

#### How to Verify:

Run the new examples `mixture_gaussian.py` and `mixture_gaussian_map.py`. See that they work.

#### Side Effects:

N/A

#### Documentation:

Doc is added in the model examples.

#### Reviewer Suggestions: 

@diengadji


dustinvtran:
I haven't done it yet, but it would great to add a custom `print_progress()` function in these examples. This would be similar to how the Bayesian linear regression examples show the distribution of lines converging around the points. Here, we can visualize the individual Gaussian components moving around and capturing the relevant data points during inference.

dustinvtran:
@diengadji: I just finished an extension from this branch. It builds an abstraction so that we can specify variational families with arbitrary factors. Let's merge this pull request ASAP so we can get to discussing the cooler one. :)

diengadji:
there is a divide by zero in the mixture_gaussian.py...the map works just fine

fmpr:
Hi,

First of all, congratulations on developing edward!
I was looking into some of the examples and I think the mixture of gaussians example is not correct. 
Consider a simple mixture of 2 gaussians with joint probability:
`\prod_{n=1}^N ( \lambda \N(x|\mu_1,\sigma_1) + (1-\lambda) \N(x|\mu_2,\sigma_2) )`
Taking the log of this gives:
`\sum_{n=1}^N \log ( \lambda \N(x|\mu_1,\sigma_1) + (1-\lambda) \N(x|\mu_2,\sigma_2) )`
However, the example code is computing (correct me if I'm wrong...):
`\sum_{n=1}^N ( \log \lambda + \log \N(x|\mu_1,\sigma_1) + \log (1-\lambda) + \log \N(x|\mu_2,\sigma_2) )`
which is not the same thing. It completely bypasses the log-sum-exp issue, which makes things easier to compute, but then the results are wrong.

I did some tests with a very simple artificial dataset consisting of a mixture of two bi-dimensional gaussians with centers on [-1,-1] and [1,1]:

`def build_toy_dataset(N):

	pi = np.array([0.4, 0.6])
	mus = [[1,1],[-1,-1]]
	stds = [[0.1,0.1],[0.1,0.1]]
	x = []
	colors = cm.rainbow(np.linspace(0, 1, 2))
	for n in xrange(N):
		k = np.argmax(np.random.multinomial(1, pi))
		x.append(np.random.multivariate_normal(mus[k],np.diag(stds[k])))
		plt.scatter(x[n][0], x[n][1], color=colors[k])
	plt.show()
	return {'x': np.array(x).astype(np.float32)}`

The means inferred by the example code are completely off:
`mean: 
[ 0.48774096 -2.39035225 -1.22405362  0.62153614]`

I changed the log lik in the example code (naively) to be:

		log_lik = []
		n_samples = get_dims(zs[0])[0]
		for s in range(n_samples):
			log_lik_z = 0.0

			for k in range(self.K):
				result = tf.exp(norm.logpdf(x, mus[s, (k*self.D):((k+1)*self.D)], sigmas[s, (k*self.D):((k+1)*self.D)]))
				result = tf.matmul(result, tf.transpose(pi))
				result = tf.reduce_sum(result, 0)
				log_lik_z += tf.log(result)

			log_lik += [log_lik_z]

This seems to give results closer to the true means:
`[-0.83513373 -1.56447959  0.8101908   0.46021864]`
However, they are still a bit off, and this is not a very numerically stable way of computing log-sum-exps. I'm new to TensorFlow... Probably this could be done in a much better way... Any tips would be appreciated.

XiaoSX:
import Error no module named edward.stats

dustinvtran:
@XiaoSX | hi, this is an an old example that's no longer supported. I recommedn seeing http://edwardlib.org/tutorials/unsupervised.

XiaoSX:
@dustinvtran Thanks, now the problem is that I got confused about the inference of the standard version of the mixture model (not collapsed one in tutorials), could you kindly do me a favor?

dustinvtran:
It's worth working out alone as an exercise. As a hint, it should be the same as the collapsed version, only now you also define an approximating distribution for the mixture assignments too.

Issue #38, opened 2016-03-07
think about what it means to "default" to reparameterization gradient


akucukelbir:
we currently default to the reparameterization gradient if the `Variational` class implements `reparam`

however, if the `Inference` class does not support reparameterization gradients (e.g. `KLpq`) then it doesn't matter whether the `Variational` class implements it or not.

PR #62, opened 2016-05-06
Feature/variational models

dustinvtran:
#### Summary:

Issues: makes progress for #7, #38

This restructures the variational modeling interface to allow a combination of variational families. Its syntax is similar to Keras and PyMC. For example, to build a mean-field variational family for the Mixture of Gaussians example:

```{Python}
variational = Variational()
variational.add(Dirichlet(1, K))
variational.add(Normal(K*D))
variational.add(InvGamma(K*D))
```

In examples where a single variational family suffices, the user can simply write
```{Python}
variational = Normal(5)
```
to do variational inference with a 5-dimensional fully factorized Gaussian. 

The foresight required here is how this extends to 

1. hierarchical variational models where we may place priors over (a subset of) variational parameters;
2. inference networks where we parameterize (a subset of) them;
3. automated computation that takes advantage of structure in certain variational families such as reparameterization for Gaussian distributions. 

I have a general but not concrete idea of how this interface will extend to them.

#### Intended Effect:

A newer and better interface for variational models.

#### How to Verify:

Run the examples and unit tests. See that they still work.

#### Side Effects:

N/A

#### Documentation:

N/A

#### Reviewer Suggestions: 

Thoughts from others is very much appreciated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,8 +33,6 @@
 *.xdy
 *.tdo
 *.pyc
-archives/
-nips2015/R/
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
@@ -61,12 +59,6 @@ var/
 *.egg-info/
 .installed.cfg
 *.egg
-
-# PyInstaller
-#  Usually these files are written by a python script from a template
-#  before PyInstaller builds the exe, so as to inject date/other infos into it.
-*.manifest
-*.spec
 
 # Installer logs
 pip-log.txt
@@ -95,9 +87,6 @@ docs/_build/
 # PyBuilder
 target/
 
-.DS_Store
-datasets/OMNI/chardata.mat
-datasets/OMNIGLOT/chardata.mat
-
-src/datasets
+examples/data/mnist
+img/
 *.ipynb_checkpoints/
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ which can be found in [`examples/`](examples/):
 
 * [Bayesian linear regression](examples/bayesian_linear_regression.py)
 * [Hierarchical logistic regression](examples/hierarchical_logistic_regression.py)
-* [Mixture of Gaussians](examples/mixture_gaussian.py)
+* [Mixture model of Gaussians](examples/mixture_gaussian.py)
 * [Gaussian process classification](examples/gp_classification.py)
 * [Bayesian neural network](examples/bayesian_nn.py)
 * [Variational auto-encoder](examples/convolutional_vae.py)
--- a/edward/data.py
+++ b/edward/data.py
@@ -37,18 +37,21 @@ class Data:
             raise
 
     def sample(self, n_data=None):
+        # TODO scale gradient and printed loss by self.N / self.n_data
         if n_data is None:
             return self.data
 
         counter_new = self.counter + n_data
         if isinstance(self.data, tf.Tensor):
             if counter_new <= self.N:
-                minibatch = self.data[self.counter:counter_new]
+                minibatch = tf.gather(self.data,
+                                      list(range(self.counter, counter_new)))
                 self.counter = counter_new
             else:
                 counter_new = counter_new - self.N
-                minibatch = tf.concat(0, [self.data[self.counter:],
-                                          self.data[:counter_new]])
+                minibatch = tf.gather(self.data,
+                                      list(range(self.counter, self.N)) + \
+                                      list(range(0, counter_new)))
                 self.counter = counter_new
 
             return minibatch
--- a/edward/inferences.py
+++ b/edward/inferences.py
@@ -125,7 +125,7 @@ class MFVI(VariationalInference):
         VariationalInference.__init__(self, *args, **kwargs)
 
     def initialize(self, n_minibatch=1, score=None, *args, **kwargs):
-        # TODO if score=True, make MFGaussian do sess.run()
+        # TODO if score=True, make Gaussian do sess.run()
         """
         Parameters
         ----------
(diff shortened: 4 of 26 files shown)


akucukelbir:
This is an ambitious step forward for Edward. I am very excited about it. 

Here are my thoughts. 

#### Interface

**1 |** To construct a variational model, we ask that the user pay very close attention to how the latent variables are used in the model. Specifically, ordering matters. Clearly, dimensions should match as well.

If we misspecify ordering or dimension, we get a bunch of `nan`'s and a
```
ValueError: Domain error in arguments.
```

I think we need to be a bit more robust here. One idea is to return with the error the ordering and the dimension of what the model expects. At the moment, we cannot give variable names, but just reporting the dimension should help.

Another idea is to also check range validities. Can we test whether we can evaluate `log_prob_zi` and report issues before starting the algorithm? For instance, what if the model requires a `Kx1` variational distribution with support over `[0,1]` (like a Beta) and we give it a `Kx1` Gamma.

**2 |** How does this interface to the other modeling languages? For instance a Stan model: are the latent variables stored in a similar fashion to a Tensorflow model?

----

I think these are important points even for research; we'll be chasing bugs in the wrong places unless we can confidently match variational models to data models.

----

#### Internal Design

Let's focus on `variationals.py`.

We now have two top level classes: `Variational` and `Likelihood`. (Presumably, we will introduce `Prior` when we begin work on hierarchical variational models.)

All the other families then inherit `Likelihood` and we have the option of either using this "raw" form of a family or a `Variational` (stacked) alternative. (See the next section for more on this.)

I would prefer to have only one object that faces the rest of the library: `Variational`. A single layer / simple variational model would then just have a single `Likelihood` and no `Prior`. 

This has many advantages: we can encode all logic that interfaces with complicated variational families in the `Variational` class. E.g. which parts to use reparameterization, which parts to apply control variates to, etc.

To accomplish this, we need to make the `Variational` class (almost) as efficient as using the "raw" variational model. This connects to the next section.

#### Performance

Right now there's an order of magnitude difference between two (mathematically) equivalent variational models.

Running `gaussian_notstacked.py`: 1.96s
Running `gaussian_stacked.py`: 15.02s

`gaussian_notstacked.py`
```{Python}
import edward as ed
import tensorflow as tf
from edward.stats import norm
from edward.variationals import Gaussian
from edward.util import get_dims

class GaussianModel:
    def __init__(self, mu, std):
        self.mu = mu; self.std = std; self.num_vars = 1

    def log_prob(self, xs, zs):
        return tf.pack([norm.logpdf(z, self.mu, self.std)
                        for z in tf.unpack(zs)])

ed.set_seed(42)
mu = tf.constant(1.0)
std = tf.constant(0.1)
model = GaussianModel(mu, std)

variational = Gaussian(model.num_vars)

inference.run(n_iter=1000)
``` 

`gaussian_stacked.py`
```{Python}
import edward as ed
import tensorflow as tf

from edward.stats import norm
from edward.variationals import Variational, Gaussian
from edward.util import get_dims

class GaussianModel:
    def __init__(self, mu, std):
        self.mu = mu; self.std = std; self.num_vars = 1

    def log_prob(self, xs, zs):
        return tf.pack([norm.logpdf(z, self.mu, self.std)
                        for z in tf.unpack(zs)])

ed.set_seed(42)
mu = tf.constant(1.0)
std = tf.constant(0.1)
model = GaussianModel(mu, std)

variational = Variational()
variational.add(Gaussian(model.num_vars))

inference.run(n_iter=1000)
```

dustinvtran:
Great points! I will get back to this. 

Regarding performance, this is a result of using the score function estimator for everything when using the stacked `Variational` class. I haven't found a way to take advantage of reparameterization for a subset of variables in this stacked class yet. My guess is that if the same experiment is done on the Bernoulli toy example then the performance would be the same (although we should check this).

dustinvtran:
#### Interface

>One idea is to return with the error the ordering and the dimension of what the model expects. At the >moment, we cannot give variable names, but just reporting the dimension should help.
>
>Another idea is to also check range validities. Can we test whether we can evaluate log_prob_zi and >report issues before starting the algorithm? For instance, what if the model requires a Kx1 variational >distribution with support over [0,1] (like a Beta) and we give it a Kx1 Gamma.

Great ideas. The first is hard to do because it requires that every model a user writes says the number of latent variables (in the examples in the repo, many cases the number is only implicitly available through the log_prob evaluation. 

Range validity would require additional metadata on the distributions to know what support they live on. This one is easy to add to the corresponding variational class, and we should do that because it's all implemented in the backend. __[TODO before merging]__

>How does this interface to the other modeling languages? For instance a Stan model: are the latent >variables stored in a similar fashion to a Tensorflow model?

Yes. There's an implicit ordering that is used and is the same for all Stan, NumPy, and TensorFlow models. So this works for the other modeling languages too.

#### Internal Design

I agree. Let's stick with interfacing purely via `Variational()`. So for simple models, while longer, it will be
```{Python}
variational = Variational()
variational = variational.add(Gaussian(model.num_vars))
```

#### Performance

I checked it on the Bernoulli toy example and it is as I mentioned. We need to figure out how to intelligently deal with reparameterizable distributions. __[TODO before merging]__

dustinvtran:
@akucukelbir: Okay, I 

+ renamed Gaussian to Normal (to follow SciPy convention);
+ changed all examples to use the interface `variational = Variational(); variational.add(...)`;
+ let `Variational()` use the reparameterization gradient if all distributions are reparameterizable, so that our current fully factorized Gaussian examples can still leverage reparameterization

More work needs to be done for

+ error checking when adding layers to `Variational()`;
+ automating different components in `Variational()`, e.g., for a mix of reparameterization and score function gradients depending on the distributions.

I suggest that we merge this pull request as it is and put the latter work as Github issues for further extensions.

akucukelbir:
@dustinvtran splendid! looks phenomenal on my end. i like all the changes and i agree that we can handle the outstanding issues later.

PR #72, opened 2016-05-07
Feature/analytic kl

### LINKED ISSUE #12: automate choice of loss function
for mean-field variational inference:
+ [x] score function gradient (general form)
+ [x] reparameterization gradient (general form)
+ [x] score function gradient when H(q) is analytic 
+ [x] reparameterization gradient when H(q) is analytic 
+ [ ] score function gradient when KL(q(z) || p(z)) is analytic 
+ [ ] reparameterization gradient when KL(q(z) || p(z)) is analytic 

(I don't know how to tell the algorithm when KL(q(z) || p(z)) is analytic, or generally any portions of the generative model.)

dustinvtran:
#### Summary:

Issues: fixes #12

This enables mean-field variational inference with an analytic KL term if all distributions added to `Variational()` are normal, and if the model has a `log_lik()` method (in which it assumes the model prior is standard normal). It is available for both reparameterization and score function gradient estimators.

This also makes progress for refactoring the `VAE` code into `MFVI`.

Possible extensions are
+ The model prior is normal but not standard normal.
+ Only part of the probability model has a normal prior and/or only part of the variational model is normally distributed. (This goes into how to leverage structure per variational family in `Variational()`.)

I suggest we add these to the Github issues after merging as they require non-trivial design decisions. This pull request also fixes `build_score_loss_entropy()` even though we never use it and generalizes `kl_multivariate_normal()`.

#### Intended Effect:

Variational inference with analytic KLs.

#### How to Verify:

Run the following. Also run the example script in `examples/bayesian_nn_analytic_kl.py`.

```{Python}
import edward as ed
import tensorflow as tf

class NormalPosterior:
    """
    p(x, z) = p(z) = p(z | x) = Normal(z; 0, 1)
    """
    def log_lik(self, xs, zs):
        return tf.constant(0.0, dtype=tf.float32)

ed.set_seed(42)
model = NormalPosterior()
variational = Variational()
variational.add(Normal())

inference.run(n_iter=1000)
inference.run(n_iter=1000, score=True)
```

#### Side Effects:

N/A

#### Documentation:

N/A

#### Reviewer Suggestions: 

N/A


mariru:
Other possible extensions:
Analytic KL for other models with conjugate priors.

Issue #27, opened 2016-02-29
a VIBES-like modeling language


dustinvtran:
It'd be nice to have a language to specify directed acyclic graphs, so we can take advantage of Rao-Blackwellization/Markov blankets. Also it'd be nice to take advantage of the case when all full conditional distributions are an exponential family.

jluttine:
Hey, great work with Edward! Do you have any insight how difficult it would be to add support for conjugate-exponential family models (or sub-graphs) so that exact gradients could be utilized (as in variational message passing)? Would it be possible or straightforward with the current machinery? I suppose it would require that the log-density functions are written as functions of the sufficient statistics instead of the variables directly and that seems like a very big change to the current code base. Anyway, I might be interested to take a look on this if I find the time but any pointers or comments would be helpful. I can also open another issue, if you consider this to be slightly off-topic. Cheers!

dustinvtran:
Hi @jluttine: The main difficulty is in the modeling language. I'm not familiar with a language that would let us take advantage of that structure. However, once it exists it is straightforward with the current machinery in a similar way you mention. The exact details depend on how fully conditional densities are encoded in the language.

dustinvtran:
> Hi Dustin,
> 
> I read you blog today. I was struck with your amazing works related to variational Bayesian Inference.
> 
> I am also into Bayesian research but more into its applications for automating stuff. I had implemented an algorithm for learning parameters for mixed Bayesian networks. But that was in Matlab. 
> 
> I want to do that in python, does any of your library support Bayesian parameter learning for mixed networks from data? Or is it possible to do that using variational approximation some how. 
> 
> Please help.
> 
> Thanks and regards
> Aakanksha

Hi Aakanksha,

The library supports inference for arbitrary joint distributions (that is, any model which can be written as a tractable joint density of data and latent variables). This includes Bayesian networks, although I haven’t found a good Bayesian network/graphical modeling language in Python, so that we can take advantage of the Markov blanket structure. Once there is a way to obtain the Markov blanket for any individual latent variable, then we can do very fast variational inference on graphical models.

For more information see https://github.com/blei-lab/edward/issues/27. If you have bandwidth to look into that direction, then it would be huge for the graphical modeling community. Hope that helps!

Dustin

datnamer:
https://github.com/pgmpy/pgmpy Have you see this? "Python Library for Probabilistic Graphical Models"

Aakanksha-B:
[...]

PR #79, opened 2016-05-13
add vertical layers

dustinvtran:
#### Summary:

+ issues: makes progress for #7; relevant to #27

`variational.add()` enables horizontal stacking. The variational model `q(z | lambda)` is a product of distributions, and the method adds another distribution as part of this product.

This feature request adds `variational.layer()`, which enables vertical stacking. Calling `variational.layer()` declares that a new layer has been added. We can place a prior `q(lambda)` on the previous layer `q(z | lambda)`. We can also place a mapping `phi(x) = lambda` to newly parameterize the previous layer `q(z | lambda)` with functions such as inference networks.

Here is the typical variational model used in variational auto-encoders. It is a mean-field normal parameterized by a neural network.
```{Python}
variational = Variational()
variational.add(Normal(model.num_vars))
variational.layer()
variational.add(mapping)
```
`mapping` is a user-defined function which takes data `x` as input and outputs the parameters of the mean-field normal. It can be simple such as constraining the set of variational parameters to be the same value, or tie all local parameters through parameters of a neural network.

This also makes progress for refactoring `VAE` into `MFVI`.

The above interface is designed so that it is amenable to the following extensions. We can think about how to best design them later.
+ __Hierarchical variational models.__
  For now, priors are not enabled and the vertical stacking only enables parameterizations.  As a proof of concept, here is a hierarchical variational model.
```{Python}
variational = Variational()
variational.add(Dirichlet(1, model.K))
variational.add(Normal(model.K*model.D))
variational.add(InvGamma(model.K*model.D))
variational.layer()
variational.add(InvGamma(model.K)) # prior on Dirichlet's alpha
variational.add(Normal(2*model.K*model.D)) # prior on Normal's mu and sigma
variational.add(InvGamma(model.K*model.D*2)) # prior on InvGamma's alpha and beta
```
  The likelihood is a typical mean-field distribution for inferring a mixture of Gaussians.  The variational prior is over the mean-field parameters.
+ __Leverage individual distribution structures.__
  This is amenable as it preserves the internal ordering of the list.
+ __More error checking.__
  Currently we want to verify that the ordering is specified
  correctly. With vertical layers we also want to verify that the node
  it is placed over matches dimensions and also that there are no
  empty layers (e.g., user can't stop after `Variational.layer()`).
+ __Invariant to the order of horizontal layers.__
  Ordering matters. We can make this invariant to the ordering with
  variable naming.
+ __Mappings and priors over combinations of variational families.__
  This is currently restricted to stacking the specific distribution
  in the list of variational families. In general we'd like to place
  parameterizations and priors over combinations of variational
  families. This is also possible with variable naming.
+ __Mappings and priors over a subset of parameters in a variational
  family and also for variational families with multiple sets of
  parameters.__
  This is currently possible although the interface is not intuitive,
  as it requires a wrapper class over multiple variational families or the use of identity functions.
  You can see that this is needed in the mixture of Gaussians example
  as we place a prior over the mean-field Gaussian's mean and standard
  deviation (they have different support).
+ __Beyond lego blocks.__
  This currently restricts hierarchical variational
  models to stacking directly on top. Similar to Keras we can generalize this to DAGs.
+ __A graphical modeling language.__
  We can use the same construction for building a graphical
  modeling language for specifying data models.

Remarks.
+ This is opposite to Keras' behavior, where `model.add()` adds vertical
  layers, and `Merge()` composes layers horizontally. Keras is
  interested in making simple horizontal layers and composing many
  vertical layers. We're interested in composing both many horizontal
  layers and many vertical layers. Thus `Merge()` is not a good
  solution.
+ This may be confusing if we put a prior on a mapping's parameters.
  The prior is a third layer but if we think in terms of hierarchical
  models, hierarchies are according to new distributions; that is,
  parameterizations are just parameterizations and priors are what
  create the next layer. Maybe this is okay though because when
  visualizing this as a graphical model we can have a deterministic
  node as the second layer and a random node as the third layer.

#### Intended Effect:

Variational models can now be globally parameterized in order to amortize inference.

#### How to Verify:

See the new example file `examples/mixture_gaussian_mapping.py`. There's no reason why I did what I did; it's just to show we can.

#### Side Effects:

N/A

#### Documentation:

Documentation is added to `Variational()` to show how this works internally.

#### Reviewer Suggestions:

@akucukelbir

### DIFF
--- a/edward/variationals.py
+++ b/edward/variationals.py
@@ -6,76 +6,136 @@ from edward.stats import bernoulli, beta, norm, dirichlet, invgamma
 from edward.util import Variable
 
 class Variational:
-    """A stack of variational families."""
-    def __init__(self, layers=[]):
-        self.layers = layers
-        self.num_vars = 0
-        self.num_params = 0
-        self.is_reparam = True
+    """
+    A stack of variational families and their parameterizations.
+    It is represented as a directed acyclic graph, where a node is a
+    variational family or parameterization. An edge i -> j exists if
+    node i is a prior on the parameters of node j or it is a
+    parameterization of node j.
 
-    def add(self, layer):
+    Parameters
+    ----------
+    graph : list
+        List of lists where the ith list is the ith layer, and each
+        element in the list (node in a layer) is a variational family
+        or mapping. The first layer can contain only variational
+        families.
+
+    Notes
+    -----
+    Currently, nodes are only connected to the node directly above it.
+    """
+    def __init__(self, graph=[[]]):
+        self.graph = graph
+        self.num_vars = 0 # num of posterior latent variables
+        self.num_params = 0 # num of variational parameters
+        self.is_reparam = True # if the variational family is reparameterizable
+
+    def add(self, node):
         """
-        Adds a layer instance on top of the layer stack.
+        Add a node instance horizontally.
 
         Parameters
         ----------
-            layer: layer instance.
+        node : variational family or parameterization
         """
-        self.layers += [layer]
-        self.num_vars += layer.num_vars
-        self.num_params += layer.num_params
-        self.is_reparam = self.is_reparam and 'reparam' in layer.__class__.__dict__
+        self.graph[-1] += [node]
+        self.is_reparam = self.is_reparam and 'reparam' in node.__class__.__dict__
+
+        if len(self.graph) == 1:
+            if not isinstance(node, Likelihood):
+                raise
+
+            self.num_params += node.num_params
+            self.num_vars += node.num_vars
+        else:
+            connected_node = self.graph[-2][self.horizontal_index]
+            if hasattr(node, '__call__'):
+                mapping = Mapping(node, connected_node.num_params)
+                connected_node.mapping = mapping
+                self.num_params += mapping.num_params
+            elif isinstance(node, Likelihood):
+                # For now this does nothing
+                raise
+            else:
+                raise
+
+            self.num_params -= connected_node.num_params
+            self.horizontal_index += 1
+
+    def layer(self):
+        """Declare a new layer. Adding now adds to this layer."""
+        self.graph += [[]]
+        self.horizontal_index = 0 # to keep track when adding to a layer
 
     def mapping(self, x):
-        return [layer.mapping(x) for layer in self.layers]
+        return [node.mapping(x) for node in self.graph[0]]
 
+    # The following methods all deal with only the first layer.
     def set_params(self, params):
-        [layer.set_params(params[i]) for i,layer in enumerate(self.layers)]
+        [node.set_params(params[i]) for i,node in enumerate(self.graph[0])]
 
     def print_params(self, sess):
-        [layer.print_params(sess) for layer in self.layers]
+        [node.print_params(sess) for node in self.graph[0]]
 
     def sample_noise(self, size):
-        eps_layers = [layer.sample_noise((size[0], layer.num_vars))
-                      for layer in self.layers]
-        return np.concatenate(eps_layers, axis=1)
+        eps_list = [node.sample_noise((size[0], node.num_vars))
+                    for node in self.graph[0]]
+        return np.concatenate(eps_list, axis=1)
 
     def reparam(self, eps):
-        z_layers = []
+        z_list = []
         start = final = 0
-        for layer in self.layers:
-            final += layer.num_vars
-            z_layers += [layer.reparam(eps[:, start:final])]
+        for node in self.graph[0]:
+            final += node.num_vars
+            z_list += [node.reparam(eps[:, start:final])]
             start = final
 
-        return tf.concat(1, z_layers)
+        return tf.concat(1, z_list)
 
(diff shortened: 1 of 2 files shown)


dustinvtran:
Conversed with @akucukelbir on Slack.

Currently the above provides "lego block"-type models which can only stack left, right, or above a node (distribution). To build a graph in general which, e.g., can stack nodes diagonally, or connect a node at layer 10 to a node at layer 2, or connect two nodes to one node, I see there being two different approaches:

1. __Variable naming.__ When adding a node (distribution), it must be given a unique character string as its name, and a character string as the node it is connected to. E.g., `variational.add(Normal(5, "x", "y"))` where this `Normal` node of 5 dimensions is labeled "x" and is connected to "y".
2. __Functional.__ All nodes (distributions) are composable. This is difficult to describe without seeing it in action; see Keras' functional API guide below on multiple-input and multiple-output models.

Relevant library design decisions:
+ [Keras uses a functional API](http://keras.io/getting-started/functional-api-guide/), which [they switched to](https://github.com/fchollet/keras/issues/1960) after not enjoying a [variable naming API](https://github.com/fchollet/keras/issues/302).
+ [Torch's `nngraph` uses a functional API](https://github.com/torch/nngraph).
+ [Lasagne uses a functional API](https://github.com/Lasagne/Lasagne).
+ [BUGS](http://www.stat.columbia.edu/~gelman/bugsR/software.pdf) and [JAGS](http://rstudio-pubs-static.s3.amazonaws.com/15236_9bc0cd0966924b139c5162d7d61a2436.html) use a variable naming API. It is based on writing the generative process and tagging variable names by the LHS of the `~`. 

I'm currently partial to Keras' solution. It has two different ways to specify the model: the easy lego block-type class via just two commands (as here in the PR), and the more general functional version. 

I find the variable naming too verbose. It only works well for BUGS and JAGS because the variable naming is baked into the generative process. But writing generative processes as the modeling language isn't useful for us where the nodes can be either deterministic or stochastic (deterministic as in a hidden unit in a neural network; stochastic as in a random variable).

dustinvtran:
from discussion with @heywhoah

dustinvtran:
__Container object__ has a list that stores all distributions.

It has a method such as `sample_params()` written in the board above. This will return a sample of all parameters that are random variables in the hierarchical model. `p` here can be another distribution in which we do `p.sample()`, a tensor object based on a function application such as `neural_network(x)` (as in recognition networks) or a `neural_network(y)` in which it is a function of another latent variable `y`. To avoid the complication where B and C have parameters which depend on the same distribution A (as in the graph above, we do this sampling procedure in order from the top layer to the bottom layer. This method will be used before any `log_prob` calculations or `sample` occurs in the variational model.

__Distribution object__ has parameters that by default are `tf.Variables()`, and they can also be set to other distribution objects. E.g., in the board above:
```{Python}
mu_z = Normal(5)
sigma_z = InvGamma(5)
z = Normal(loc=mu_z, scale=sigma_z)
```
`mu_z` and `sigma_z` are distribution objects with their parameters set as `tf.Variables()` to be optimized freely (location and scale; shape and scale respectively). `z` is a normal distribution with location and scale parameters set to other distributions. To sample from `z` requires first realizing its parameters. This is done recursively as described above in `sample_params()`. To calculate a joint density `log q(z, mu_z, sigma_z)` is easy based on `log q(z | mu_z, sigma_z) + log q(mu_z) + log q(sigma_z)`. 

This also applies to inference networks. For example,
```{Python}
def neural_network(x):
    return tf.matmul(tf.tanh(tf.matmul(x, W_1) + b_1), W_2) + b_2

NNx = neural_network(x)
z = Normal(loc = NNx)
```
`z` is a normal distribution with scale parameter as a free parameter to optimize over. To sample from `z` requires first realizing its location parameter. This is also done recursively as described above in `sample_params()` which to realize `NNx` first requires realizing which data point `x`.

Interface-wise, Matt's proposal is to do something like PyMC3, e.g.,
```{Python}
with model:
    mu_z = Normal(5)
    NNx = neural_network(x)
    z = Normal(loc=mu_z, scale=NNx)
```

+ Parameters can be one of three things: free parameters (`tf.Variables`) to be directly optimized over, random variables, or outputs of functions. 
+ Things like graph structure and Markov blankets are determined as the distributions are added to the container object. Some bookkeeping is required to do this and also things such as whether samples to sample from will leverage TensorFlow-based samplers or SciPy-based samplers.
+ Things like conjugacy, analytic entropy terms, and analytic KL are all attributes in each distribution object; some may require communication between the variational distribution and probability model (e.g., conditionally conjugate updates, analytic KL's). 
+ The inputs of the inference network/inverse mapping are determined precisely by the user when defining `NNx`, which is just a tensor object that `model` works with (it never directly works with the function). Parameter tying is just a form of inference network/inverse mapping.

dustinvtran:
I'm closing this pull request. #126 makes progress for a newer modeling language spec. It is more agreeable to all the criteria we'd like to satisfy. Matt and others at Google Brain also agree.

Issue #66, opened 2016-05-07
allow for a mixing of reparam and score function gradient estimation in Variational()


akucukelbir:
see https://github.com/blei-lab/edward/pull/62#issuecomment-217540682 and https://github.com/blei-lab/edward/pull/62#issuecomment-217578844

Issue #33, opened 2016-03-05
entropy function for distributions


dustinvtran:
references
+ https://en.wikipedia.org/wiki/Differential_entropy#Differential_entropies_for_various_distributions
+ https://github.com/scipy/scipy/blob/05dc7b8b0662c9c973203b36044458ca35084833/scipy/stats/_continuous_distns.py

altosaar:
tf.contrib.distributions has entropy functions

PR #85, opened 2016-05-15
Feature/analytic entropies

dustinvtran:
#### Summary:

+ issues: makes progress for #33 and leveraging the structure of variational models in #66

I added entropies to all distributions that are currently available in `Variational()`. This lets us leverage analytic entropies during inference. In particular, we no longer need to take stochastic gradients of the entropy during mean-field variational inference and can instead use analytic gradients of the entropy.

(This pull request also standardizes a few arguments in other unit tests and changes the names of some utility functions so that they are more standard. I'm a bit lazy to tease out these changes and push them separately.)

__Before merging, there is one problem that I need to figure out__: I'm not sure if inference is actually faster.. it seems like the examples are in fact taking longer to converge and/or run.

#### Intended Effect:

Entropies of certain distributions are available. Entropies of all variational distributions are available. Mean-field variational inference is faster.

#### How to Verify:

Run all examples. They are faster. Run all new unit tests. They work.

#### Side Effects:

N/A

#### Documentation:

Documentation is added to the `Distributions` class.

#### Reviewer Suggestions:

N/A


dustinvtran:
For now I deactivated the use of the analytic entropies during variational inference. It remains open how useful they are. I'm merging the pull request for the other features.

PR #87, opened 2016-05-18
fix Variational init for instances containing layers

meereeum:
As it stands, was throwing ```AttributeError: 'list' object has no attribute 'num_factors'``` -- the list in question being ```layers```, when it should be the elems of that list.


dustinvtran:
Thanks for this! In general, more unit testing should go into the variational class to avoid trivial coding errors like this. I'm hoping I can get to it during summer post-NIPS deadline.

PR #90, opened 2016-05-23
Feature/example mixture density network

### LINKED ISSUE #17: add support for model compositionality with keras
Seems like [keras](http://keras.io) is winning out of all the deep learning abstraction-based libraries. 

It would great to be compatible with [keras' compositionality](http://keras.io/models/) for adding neural network layers. This would be convenient for stacking NN layers following their Sequential and Graph Model classes, as options to parameterize the probability model and/or the variational model.

It looks like all we need to do is write a wrapper around their `evaluate()` method, and hack evaluation of the log-likelihood as the compiled model object's "loss".
* https://github.com/fchollet/keras/blob/master/keras/models.py
* http://blog.keras.io/keras-as-a-simplified-interface-to-tensorflow-tutorial.html

### LINKED ISSUE #88: Using Edward in Mixture Density Networks
Given that Edward implements many distributions in TF, I thought it would be a natural to use these distributions for Mixture Density Networks (MDN). 
Mixture Density Networks are a type feed forward neural net that make a prediction as mixture of  distributions. See a blog post I wrote about them : http://cbonnett.github.io/MDN.html

The combination of Edward, TF and Keras would make creating these easier. 
There seems to be one roadblock for using Edward for  MDN's and that is the following line in each of the density distributions: 
 
```python
x = tf.cast(tf.squeeze(x), dtype=tf.float32)
```

And in particular the tf.squeeze for x when dealing with a 1-D likelihood (I have yet to try higher dimensions). In Particular, TF complains on the following line : 

```python
z = (x - loc) / scale
``` 
in the ```norm ```function when ```loc``` and ```scale``` have dimension ```[? ,10]``` if we are trying to fit a mixture of 10 normal distrbutions and x has ```[?,1]``` before ```squeeze``` is applied. TF complains about the dimensionality, ```x``` needs that extra dimension for MDN's.  
 
Does Edward break if the squeeze is removed ? (i'm guessing it will)
Any idea for a solution for this ? 

Here is the Full MDN code but without Edward to the above mentioned problem:

```
from keras.layers import Input, Dense, merge
import tensorflow as tf
sess = tf.Session()
from keras import backend as K
K.set_session(sess)

mixture_components = 10
NSAMPLE = 6000

y_data = np.float32(np.random.uniform(-10.5, 10.5, (1, NSAMPLE))).T
r_data = np.float32(np.random.normal(size=(NSAMPLE,1))) # random noise
x_data = np.float32(np.sin(0.75*y_data)*7.0+y_data*0.5+r_data*1.0)

X_train, X_valid, y_train, y_valid = train_test_split(x_data, y_data, random_state=42)

print x_data.shape, X_train.shape
print y_data.shape, y_train.shape

X = tf.placeholder(tf.float32, shape=(None, 1))
y = tf.placeholder(tf.float32, shape=(None, 1))

hidden1 = Dense(25, activation='relu')(X)  # fully-connected layer with 128 units and ReLU activation
hidden2 = Dense(25, activation='relu')(hidden1)
means = Dense(mixture_components)(hidden2)
standard_deviations = Dense(mixture_components, activation=K.exp)(hidden2)
weights = Dense(mixture_components, activation=K.softmax)(hidden2)

def logpdf(x, loc=0, scale=1):
    z = (x - loc) / scale
    return (-0.5*tf.log(2*np.pi) - tf.log(scale) - 0.5*tf.square(z))

def get_lossfunc(weigts, standard_deviations, means, y):
    result = tf.exp(logpdf(y, means, standard_deviations))
    result = tf.mul(result, weights)
    result = tf.reduce_sum(result, 1, keep_dims=True)
    result = -tf.log(result)
    return tf.reduce_mean(result)

print weights.get_shape


lossfunc = get_lossfunc(weights, standard_deviations, means, y)
train_op = tf.train.AdamOptimizer().minimize(lossfunc)

sess = tf.InteractiveSession()
sess.run(tf.initialize_all_variables())

NEPOCH = 20
train_loss = np.zeros(NEPOCH) # store the training progress here.
valid_loss = np.zeros(NEPOCH)
for i in range(NEPOCH):    
    _, train_loss[i] = sess.run([train_op, lossfunc], feed_dict={X: X_train, y: y_train})
    valid_loss[i] = lossfunc.eval(feed_dict={X: X_valid, y: y_valid})
    
    pred_weights, pred_means, pred_std = sess.run([weights, means, standard_deviations], 
                                                  feed_dict={X: X_valid})
````

### PR DESCRIPTION
**Pull request by dustinvtran**
#### Summary:

+ issues: fixes #88; fixes #17; makes progress for #66; relevant to #51

This makes progress for enabling idiomatic usage of TensorFlow with Edward, adds another model example, and makes progress for the VAE refactor.

+ `tf.Variables()` can now be instantiated in two locations:

  1. when defining variational models;
  2. optionally, when the user defines the probability model.

  For any `tf.Variables()` that the user defines within the probability model, they will be inferred as point estimates rather than as distributions. This defines a "EM" where 1 can be roughly thought as being inferred with the "E-step" and 2 as the "M-step". We do this because 2 is explicitly defined as parameters by the user. For example, when specifying neural networks in probability models with Keras or PrettyTensor, the wrappers define their own TensorFlow variables. It is difficult if not impossible to be Bayesian about their parameters.

  In terms of joint distributions, it is now something like `p(x, (z,theta))`, where `x` is data, `z` are latent variables to be Bayesian about (which we define variational models to infer their posterior), and `theta` are "model parameters" to be point estimated.  The user still defines a joint density `p(x, z)` where `z` are taken as inputs.  However, `theta` are implicitly taken as inputs when the user defines them; see the MDN example why this is necessary.

  Note a nuance with `ed.MAP`, in which the latent variables `z` we define are also point estimated. However, these `z`'s are more general than the `theta`s in that we can choose to be more Bayesian about them by running a different inference algorithm. `theta` will always be point estimated.

  Internally, there is no change to the code.
+ Samples from the variational model can be a mix of tensors and np.arrays. This lets us leverage TensorFlow-based samplers, which do not require realizing variational parameters as it does for SciPy-based samplers. It speeds up computation. It is also automated in that samples will be a Tensor or np.array based on each variational factor. It is one step towards leveraging distribution structure during inference.

  Motivating examples:
  + `sample_noise()` in `Normal` now returns a tensor.
  + `sample()` in `PointMass` now returns a tensor.

  Internally, this is how it works:
  + In variational factors:
    `sample_noise()` and `sample()` return either a tensor or np.array.
  + In Variational:
    `sample()` is the main method. It returns samples and also a list of samples, where each list element corresponds to samples from a variational factor. If not a tensor, the samples will be a placeholder for being fed a sampled np.array during the session. `np_sample()` fills in the placeholders.
  + In Inference:
    We feed np.ndarray's to any placeholders in `self.samples` during `inference.update()` by calling `variational.np_sample()` to form the dictionary.
+ data can be a list of placeholders, as part of feed_dict and fed np.arrays during inference. This is the most popular way for handling data with TensorFlow. This is useful as in the MDN example.  data can also be a list of np.ndarrays.
+ Added mixture density network as an example. It showcases Keras, data given by placeholders, and generally Edward in a more idiomatic TensorFlow usage.
+ `sess` can be given by user and then handed to the inference. This is helpful as in the MDN example.
+ `PointMass` can now be used as a variational factor, e.g., during `variational = Variational(); variational.add(PointMass(...))`
+ use `self.loss` instead of `self.losses`

Note this breaks the `mixture_gaussian.py` example. This is unrelated to the code changes and more the fact that inference with inverse-Gamma samplers is really hard. It blows up very quickly; I haven't found a robust solution yet.

#### Intended Effect:

See above.

#### How to Verify:

Run all examples and unit tests.

#### Side Effects:

N/A

#### Documentation:

N/A

#### Reviewer Suggestions:

N/A

### CODE REVIEW

**Comment by akucukelbir** [comment]
initial comment: keras automatically defaults to a Theano backend.

changing this either 
1. modifying a local config file: `~/.keras/keras.json`
2. declaring an environment variable: `KERAS_BACKEND=tensorflow`

is there a way of doing this locally: just make `example/mixture_density_network.py` run with tensorflow, regardless of the user's default preference?

File: edward/data.py [hunk 1]
```diff
@@ -30,57 +38,71 @@ def __init__(self, data=None, shuffled=True):
             # shuffle self.data
             raise NotImplementedError()
 
-        self.counter = 0
         if self.data is None:
-            self.N = None
+            pass
         elif isinstance(self.data, tf.Tensor):
             self.N = self.data.get_shape()[0].value
+            self.counter = 0
```
akucukelbir: @dustinvtran can you document what counter is? i don't intuitively expect a counter in a data class, and it doesn't pass the 60 second test. (i'm still trying to parse what the counter is doing.)

edward/data.py
```diff
@@ -30,57 +38,71 @@ def __init__(self, data=None, shuffled=True):
...
         elif isinstance(self.data, np.ndarray):
             self.N = self.data.shape[0]
+            self.counter = 0
+        elif isinstance(self.data, list):
+            if isinstance(self.data[0], np.ndarray):
+                self.N = [x.shape[0] for x in self.data]
+                self.counter = [0]*len(self.data)
+            else: # list of placeholders
```
akucukelbir: placeholders for what?

edward/data.py
```diff
@@ -30,57 +38,71 @@ def __init__(self, data=None, shuffled=True):
...
         elif isinstance(self.data, dict):
             pass
         else:
-            raise
+            raise NotImplementedError()
 
     def sample(self, n_data=None):
         # TODO scale gradient and printed loss by self.N / self.n_data
-        if n_data is None:
+        if n_data is None or self.data is None:
```
akucukelbir: i'm not sure i follow this TODO

edward/data.py
```diff
@@ -30,57 +38,71 @@ def __init__(self, data=None, shuffled=True):
...
+                        counter_new = counter_new - self.N[i]
+                        minibatch[i] = np.concatenate((self.data[i][self.counter[i]:],
+                                                       self.data[i][:counter_new]))
 
-            return minibatch
+                    self.counter[i] = counter_new
+
+                return minibatch
+            else: # list of placeholders
+                raise NotImplementedError()
```
akucukelbir: consider adding these comments as strings within the exception

edward/data.py
dustinvtran: `tf.placeholder`

edward/data.py
akucukelbir: i now see that counter wraps around the axis that is being subsampled... perhaps some jargon that involves the words `window` or `sliding` and perhaps the word `index` might be more intuitive?

dustinvtran: you have a big data set of size `N`. `counter` tells us where in this data set we currently are to get the next mini-batch of size `n_data`, i.e., data indices `self.counter:(self.counter + n_data)` with the nuance of looping back when `self.counter + n_data > N`.

i will document this

edward/data.py
dustinvtran: this is a TODO unrelated to this PR; for anything that does data subsampling we technically need to scale the log-likelihood evaluation by that factor. there is no current way to do this (also a problem in ADVI in Stan)

examples/mixture_density_network.py
```diff
@@ -5,7 +5,7 @@
 Probability model:
     Prior: None ("flat prior")
     Likelihood: Mixture sum of normals parameterized by a NN
-Inference: Maximum a posteriori
+    Inference: Maximum a posteriori
```
dustinvtran: Isn't inference separate from the probability model?

edward/data.py
akucukelbir: roger that. still see my note above re jargon

edward/data.py
akucukelbir: we took care of this in stan, no? (not elegantly though)

examples/mixture_density_network.py
```diff
@@ -5,7 +5,8 @@
 Probability model:
     Prior: None ("flat prior")
     Likelihood: Mixture sum of normals parameterized by a NN
-    Inference: Maximum a posteriori
+
```
dustinvtran: can i also say that there shouldn't be an empty line between probability model and inference/variational model? :) (we could, although for consistency we would need to do it for all the examples)

edward/data.py
dustinvtran: the current experiments do the wrong thing because it adds log(scale factor) rather than scale factor * log_likelihood_on_minibatch + log_prior

akucukelbir: i'm confused. i thought i fixed that months ago: https://github.com/stan-dev/stan/blob/adsvi/how_to_ADSVI.md

edward/models/variationals.py
```diff
@@ -535,8 +591,16 @@ def print_params(self, sess):
...
-        return tf.reshape(self.params, [1, self.num_params])
+        # parameter samples (as in black box variational methods).
+        return tf.pack([self.params]*size)
+
+    def log_prob_zi(self, i, zs):
+        """log q(z_i | lambda)"""
+        if i >= self.num_factors:
+            raise IndexError()
+
+        return tf.cast(tf.equal(zs[:, i], self.params[i]), dtype=tf.float32)
```
akucukelbir: this is a bit mysterious.

akucukelbir:
`inferences.py` lgtm

akucukelbir:
okay. the rest looks good to me too.

summary:
1. example won't run with fresh install of keras.
2. consider different jargon around `counter` and subsampling
3. consider adding strings to exceptions
4. consider clarifying `PointMass` in `variationals.py`

otherwise, splendid!

edward/data.py
dustinvtran: yeah im referencing the paper experiments. to scale the gradient appropriately requires direct access to the split of log-likelihood + log-prior, not just the log-joint.

cbonnett:
It doesn't seem possible to set TF as default backend in Keras. I asked the Keras Slack chat, I'll let you know if anything useful comes back.  Second best thing might be to test  to see if the backend is Theano and then instruct the user how to change or refer them to instructions on how to change the backend to TF.

edward/data.py
akucukelbir: an yes. we'll redo those for the revision.

dustinvtran:
I fixed an error in the pull request which now makes `mixture_gaussian.py` work again. The problem was that for score function gradients on the normal samples, TensorFlow was being too smart and auto-diffed through the reparameterization as well. For score function gradients, this shouldn't happen.

As an example, the following two did not work in this branch. But with the fixed changes they work:
```{Python}
import edward as ed
import tensorflow as tf

from edward.stats import norm

class NormalPosterior:
    """
    p(x, z) = p(z) = p(z | x) = Normal(z; mu, std)
    """
    def __init__(self, mu, std):
        self.mu = mu
        self.std = std

    def log_prob(self, xs, zs):
        return norm.logpdf(zs, self.mu, self.std)

ed.set_seed(42)
mu = tf.constant(1.0)
std = tf.constant(1.0)
model = NormalPosterior(mu, std)
variational = Variational()
variational.add(Normal())

inference.run(n_iter=10000, score=True)
```
```{Python}
import edward as ed

from edward.models import Variational, Bernoulli, Normal
from edward.stats import bernoulli, norm

class NormalBernoulliPosterior:
    def log_prob(self, xs, zs):
        return norm.logpdf(zs[:,0], 1.0, 1.0) + \
               bernoulli.logpmf(zs[:,1], 0.6)

ed.set_seed(42)
model = NormalBernoulliPosterior()
variational = Variational()
variational.add(Normal())
variational.add(Bernoulli())

inference.run(n_iter=10000)
```

Also to make `mixture_gaussian.py` more robust, I added some numerical thresholds on inverse gamma sampling and its unconstrained to constrained transformation.

dustinvtran:
@akucukelbir 

1. Following @cbonnett, I guess we can add a note in the wiki documentation somewhere, that in order to use Keras in Edward you need to set Keras' backend to TensorFlow. I'm less inclined to put this note in the `mixture_density_network.py` file explicitly; this is because I expect we'll have many examples in the future using Keras and it would be too verbose to have that note everywhere.
2. Added doc.
3. This can be done when we improve our error checking as part of another feature. That is, all raised errors at the moment have no message so we can change all of them at once.
4. Added a short line explaining how that `log_prob_zi` works.

I will now merge this pull request.

Sandy4321:
good, thanks

PR #91, opened 2016-05-23
simplify plotting model examples

dustinvtran:
This is a simple clean-up of the model examples which plot things during inference. It is meant to be easier to read. Also, since `bayesian_linear_regression.py` is the first example shown on the front page, I removed plotting from it because plotting makes the first example needlessly complex. The example with plotting is still available at `bayesian_linear_regression_plot.py`. 

I would merge this but I'm having weird issues with matplotlib not working in virtualenv (and now suddenly not working on root either). Would be good if someone could double check that the plots all still work.

### DIFF
--- a/examples/bayesian_linear_regression_plot.py
+++ b/examples/bayesian_linear_regression_plot.py
@@ -11,6 +11,7 @@ Variational model
 """
 import edward as ed
 import tensorflow as tf
+import matplotlib.pyplot as plt
 import numpy as np
 
 from edward.models import Variational, Normal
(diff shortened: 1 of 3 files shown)


akucukelbir:
okay. i think tensorflow is being too smart here.

this thing below works for me.

i create a new figure each time `print_progress` is called and the modulo operation evaluates to `True`. no window remains after completion of the script.

```{Python}
def print_progress(self, t, losses, sess):
    if t % self.n_print == 0:
        print("iter %d loss %.2f " % (t, np.mean(losses)))

        # Sample functions from variational model
        mean, std = sess.run([self.variational.layers[0].m,
                              self.variational.layers[0].s])
        rs = np.random.RandomState(0)
        zs = rs.randn(10, self.variational.num_vars) * std + mean
        zs = tf.constant(zs, dtype=tf.float32)
        inputs = np.linspace(-8, 8, num=400, dtype=np.float32)
        x = tf.expand_dims(tf.constant(inputs), 1)
        mus = tf.pack([self.model.mapping(x, z) for z in tf.unpack(zs)])
        outputs = sess.run(mus)

        # Get data
        y, x = sess.run([self.data.data[:, 0], self.data.data[:, 1]])

        # Plot data and functions
        fig = plt.figure(figsize=(8,8), facecolor='white')
        ax = fig.add_subplot(111, frameon=False)
        ax.plot(x, y, 'bx')
        ax.plot(inputs, outputs.T)
        ax.set_ylim([-2, 3])
        plt.show(block=False)
```

akucukelbir:
ah damn... this is simply creating windows on top of each other...

dustinvtran:
I added `plt.pause(1.0/60.0)` after `plt.draw()` which got it to work for me. My guess is that it was refreshing too quickly. Would be great if someone can double check that it works for them now too. If not I will use a company laptop tomorrow to confirm.

akucukelbir:
lgtm. merging.

PR #100, opened 2016-05-25
Take prediction part out of training loop.

cbonnett:
The prediction part:

```
pred_weights, pred_means, pred_std = sess.run(
        [model.pi, model.mus, model.sigmas], feed_dict={X: X_test})
```
Is not now calculated for every training iteration, took it out of the training loop.

### DIFF
--- a/examples/mixture_density_network.py
+++ b/examples/mixture_density_network.py
@@ -82,5 +82,6 @@ for i in range(NEPOCH):
                                 feed_dict={X: X_train, y: y_train})
     test_loss[i] = sess.run(inference.loss, feed_dict={X: X_test, y: y_test})
     print("Train Loss: {:0.3f}, Test Loss: {:0.3f}".format(train_loss[i], test_loss[i]))
+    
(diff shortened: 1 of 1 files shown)


examples/mixture_density_network.py
```diff
@@ -82,5 +82,5 @@ def build_toy_dataset(nsample=6000):
                                 feed_dict={X: X_train, y: y_train})
     test_loss[i] = sess.run(inference.loss, feed_dict={X: X_test, y: y_test})
     print("Train Loss: {:0.3f}, Test Loss: {:0.3f}".format(train_loss[i], test_loss[i]))
-    pred_weights, pred_means, pred_std = sess.run(
+pred_weights, pred_means, pred_std = sess.run(
```
dustinvtran: Cool. That makes sense since we don't actually print anything about these predictions during training. 

Could you add an empty line between the end of the for loop and this line?

cbonnett: Done

PR #101, opened 2016-05-26
removed second tf.sum in log_prob

cbonnett:
I removed the second ```tf.reduce_sum``` in ```log_prob``` method of ```class MixtureDensityNetwork```
I don't think it's supposed to be there.

### DIFF
--- a/examples/mixture_density_network.py
+++ b/examples/mixture_density_network.py
@@ -47,8 +47,7 @@ class MixtureDensityNetwork:
         result = tf.exp(norm.logpdf(y, self.mus, self.sigmas))
         result = tf.mul(result, self.pi)
         result = tf.reduce_sum(result, 1, keep_dims=True)
-        result = tf.log(result)
-        return tf.reduce_sum(result)
+        return tf.log(result)
 
 def build_toy_dataset(nsample=6000):
     y_data = np.float32(np.random.uniform(-10.5, 10.5, (1, nsample))).T


dustinvtran:
I believe it's required: the first `reduce_sum` sums over the number of components. Then we take `tf.log` to calculate the log-likelihood. At this point we have a vector of log-likelihoods, where each element corresponds to the log-likelihood of a data point. We then apply the last `tf.reduce_sum` because the log-likelihood over the data set is the sum of log-likelihoods over the individual data points. Feel free to correct me if I'm wrong.

dustinvtran:
Oh, I don't think we need the `keep_dims=True` for the first `reduce_sum` though.

cbonnett:
Indeed, the ```keep_dim``` is not needed. 

As for the ```reduce_sum``` here are the loss plots. 

The loss function form the pull request :
loss_plot_pull_request

The loss function that happens in current branch :
loss_plot_before_pull_request

So one is the total loss (if I'm not mistaken), while the other is average loss per datapoint. To some extent is choice I suppose. In the current notebook I have the first plot.

dustinvtran:
Right, that makes sense. 

Due to black box variational inference reasons, the output of `log_prob(xs, zs)` is expected to be a vector of joint log densities `[log_prob(xs, zs[1,:]), ..., log_prob(xs, zs[S, :])]`, where the sth element is the joint log density evaluated at the data set `xs` and the sth sample of latent variables `zs[s, :]`. In the MDN example, the joint log density is just the log-likelihood of the data set, which is the sum over log-likelihoods for each data point.

Currently, the averaging happens only by coincidence because of the line `tf.reduce_mean(self.model.log_prob(x, z))` in `MAP`. In MAP the vector always has only 1 element, so it should be the same as `tf.squeeze(self.model.log_prob(x, z))`. In fact for clarity I'll change it to `tf.squeeze()` now.

You're right that ultimately whether we indicate the loss with respect to a single data point or the full data set is a design choice. I decided on the latter because of the definition of `MAP` and its loss function. During user post-processing, we can easily change it to reflect the average though.

dustinvtran:
Changes made in https://github.com/blei-lab/edward/commit/1d3d279ec34d886be854d0a86bdaefa3901a3984

cbonnett:
Okay, great. 

I'm closing this.

Issue #104, opened 2016-05-27
Add Train/Test Separation to convolutional_vae example


jramapuram:
The current example only shows the train division. To complete the test division scoping needs to be added. I.e. `phase=pt.Phase.train` or `phase=pt.Phase.test` . The way it is currently setup is incorrect as the testing is evaluated with the train phase. This causes the following issues:
  - Dropout is evaluated on test and train
  - batch_normalization is applied on test and train the same way (test needs to not update the running mean/var)

mariru:
I think the train/test data split is required more generally. Not only for the convolutional_vae_example. We might consider adding an optional Data.test_data attribute to the Data class.

For evaluation we can then simply run Model.log_prob() on the test data.

jramapuram:
Sure, that works as well. Just wanted to bring up the inconsistency with only having training logic and no testing logic!

dustinvtran:
Sure that makes sense. Part of the reason I avoided this at the moment is due to the lack of a criticism interface. Criticism is all about assessing model fit with observed data, which in practice you do using data held-out from training time. So before we do any training/test split, we want to think hard about how to build the criticism interface following a principled probabilistic approach. Maja's pull request #105 makes sense as a first approach to this.

PR #105, opened 2016-05-27
feature/train-test-split 

mariru:
**Summary:**

    issues: makes progress for #104 Add Train/Test Separation to convolutional_vae example

I added a test_data attribute to the Data class and a test() method to the VariationalInfernce class. The test method evaluates an evaluation metric on the test data. The evaluation metric should be specified in the model (see examples/bayesian_linear_regression_multidimensional.py where the evaluation metric is MSE). If no evaluation metric is specified, test() evaluates the log probability of the model on the test data.

In addition, I added examples/bayesian_linear_regression_multidimensional.py
This is a modified version of examples/bayesian_linear_regression.py which takes multidimensional input vectors and evaluates MSE on the test data.

**Intended Effect:**

Easy evaluation on test data.

**How to Verify:**

Run examples/bayesian_linear_regression_multidimensional.py
Can, also run all older examples, they are not affected by the changes.

**Documentation:**

N/A

**Reviewer Suggestions:**

@dustinvtran

### DIFF
--- a/edward/data.py
+++ b/edward/data.py
@@ -35,8 +35,9 @@ class Data:
     is used to obtain the next batch of data starting from
     self.counter to the size of the data set.
     """
-    def __init__(self, data=None, shuffled=True):
+    def __init__(self, data=None, test_data=None, shuffled=True):
         self.data = data
+        self.test_data = test_data
         if not shuffled:
             # TODO
             # shuffle self.data
@@ -61,6 +62,7 @@ class Data:
             pass
         else:
             raise NotImplementedError()
+
 
     def sample(self, n_data=None):
         # TODO scale gradient and printed loss by self.N / self.n_data
--- a/edward/inferences.py
+++ b/edward/inferences.py
@@ -26,6 +26,7 @@ class Inference:
         self.model = model
         self.data = data
 
+
 class MonteCarlo(Inference):
     """
     Base class for Monte Carlo methods.
@@ -118,6 +119,17 @@ class VariationalInference(Inference):
 
     def build_loss(self):
         raise NotImplementedError()
+
+    def test(self,sess):
+        x = self.data.test_data
+        z, self.samples = self.variational.sample(x)
+    
+        eval_metric = getattr(self.model, 'evaluation_metric', None)  
+        if eval_metric is None:
+             eval_metric = getattr(self.model, 'log_prob', None)  
+
+        loss = sess.run([eval_metric(x, z)])
+        return loss[0][0]
 
 class MFVI(VariationalInference):
 # TODO this isn't MFVI so much as VI where q is analytic
--- a/examples/bayesian_linear_regression_multidimensional.py
+++ b/examples/bayesian_linear_regression_multidimensional.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+"""
+Bayesian linear regression using mean-field variational inference.
+
+Probability model:
+    Bayesian linear model
+    Prior: Normal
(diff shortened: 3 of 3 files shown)


dustinvtran:
As mentioned in #104, I think this is a good first approach to the training/test split. In general, this would be subsumed under the criticism interface. So we want to think carefully about how to do this so that it generalizes to things like posterior predictive checks and cross validated assessments of model fit. E.g., should testing be part of inference or its own separate object (possibly taking in the fitted model and data as input)?

jramapuram:
From a purely programatic point of view this still does not address the issues with for example:
  - use of scoping w/ prettytensor [i.e. phase is not passed in]
  - building two objects whereby the test object re-uses the underlaying train params.

dustinvtran:
Okay so I thought about this a little more. 

+ I like that a test function must take `sess` as input. The alternative is to have inference return all learned parameters as NumPy arrays, and then the user would somehow have to wield these NumPy arrays for evaluation at test time. That sounds not very feasible.

Your structure also motivates a lot of things. So here are additional proposals:

+ There will be a `criticism` object. It can do test set and cross validated evaluations, or a PPC.
+ The input is `model`, `variational`, `data` (test or training data), and `sess`. Some of these inputs are optional and more may be necessary. For example, prior predictive checks don't require any inference; therefore it can use its own session and does not have `variational` as input. Predictive checks will also require a discrepancy function `t()` as input.
+ Evaluation metrics return a scalar. PPCs return a reference distribution and realized discrepancy. Graphical checks return a plot.
+ The criticism requires a method from the `model` object. What is the model method's inputs and outputs? 

  Evaluation metrics. The inputs are the usual inputs to `log_prob()`, i.e., data `xs` and latent variables `zs`. The output is the mean of the likelihood for each set of latent variables. It is a TensorFlow vector `[ E_{p(xs | zs[1, :])} [xs], ..., E_{p(xs | zs[S, :])} [xs] ]`.

  PPC's. The inputs are the size of the replicated data set `size`, and a `S x d` matrix of latent variables `zs`. The output is a list of replicated data sets from the likelihood, `[x^{rep, 1}, ..., x^{rep, S}]`, where `x^{rep, s} ~ p(x | zs[s, :])` and `x^{rep, s}` has `size`-many data points. It is a list of objects where the type of the object is the same as the type of the test data.

Let's use your example as reference. We want to calculate mean squared error on held-out data. We will require the model to have a method for evaluation metrics as explained above. The `criticism` object does the following: 1. sample a mini-batch of latent variables from the variational distribution; 2. calculate a vector of likelihood predictions, one element for each sample of latent variables; 3. take the average of this vector, which is our Monte Carlo estimate of the posterior predictive mean; 4. calculate the MSE of this value.

Here are open questions:

+ Should criticism be a class that has a bunch of methods we can call for each criticism we want to do? Or should it be a bunch of standalone functions that we call?
+ What do we call the model method? For evaluation metrics, I think it makes sense to call it `predict()`. For PPC's, I think it makes sense to call it `sample_lik()`.
+ How can do we do criticism during inference? For example, what if want to we print out the test loss as a function of iterations in the algorithm? No clue; may for now users will have to do this in Native TensorFlow similar to the mixture density network example.

Thoughts much appreciated.

dustinvtran:
Following the above proposal, it also extends to @jramapuram's points. Namely, PrettyTensor test scoping will be inside the criticism object and we can add PrettyTensor training scoping inside the inference object. The criticism object can also can re-use the underlying train params because it has access to everything (`model`, `variational`, `data`).

PR #117, opened 2016-06-04
global session (interactive)

### LINKED ISSUE #111: internally use global sessions
It's kind of a pain to plug in all these arguments any time we want to criticize model fit. By default the functions should grab the session globally from the environment, if it exists. To do this, inference will be modified to globally instantiate the session rather than just locally.

As a reference, here's how [Keras handles it](https://github.com/fchollet/keras/blob/master/keras/backend/tensorflow_backend.py#L29).

mariru:
#### Summary:

+ issues: fixes #111 

This changes tensorflow session management. Rather than having to pass the tf session as an argument it is now instantiated as a global variable. In addition, using tf.InteractiveSession() rather than tf.Session() simplifies syntax further. 

#### Intended Effect:

Simplify edward development. Simplify the usage of inference and criticism functionality without requiring the user to manage the tf session. 

#### How to Verify:

Run examples. (Note: all current examples have been tested with global sessions)

#### Side Effects:

Not backward compatible. Many functions no longer take `sess` as an argument. So old examples may not run anymore. All edward example models in 'examples/'  have been fixed to cohere with the new tf session management.

The following substitutions make old code compatible with the new implementation:

`sess.run(t)` -> `t.eval()`
`sess.run([...])`  -> `tf.get_default_session().run([...])`
`self.update(sess)` -> `self.update()`

#### Documentation:

N/A

#### Reviewer Suggestions:

@dustinvtran


dustinvtran:
This is awesome. Sessions are annoying to handle in high-level libraries so getting it defined globally helps abstract the library away from the backend. Your mapping of `sess.run(t)` -> `t.eval()` and `self.update(sess)` -> `self.update()` makes sense.

I would also suggest defining two utility functions in `util.py`:

+ `get_session()` will do the lines as you did where
```
def get_session():
  """Get the session defined globally; if not already defined, then
  the function will create a global session."""
  global _ED_SESSION
  if tf.get_default_session() is None:
      _ED_SESSION = tf.InteractiveSession()
  return _ED_SESSION
```
+ `set_session()` will set the global session in case the user has already used the session during other TensorFlow stuff. This will let `inference.initialize()` still take `sess` as an optional argument where we do something like `if sess == None:`, `set_session(sess)`.
```
def set_session(sess):
  """Sets the global session used in Edward."""
  global _ED_SESSION
  _ED_SESSION = sess
```

Then instead of putting those `global _ED_SESSION [...]` lines at the top of the Edward files, I would write `get_session()` in any of the functions or methods that require the session. (This makes the modularization easier to tease things out.) And instead of the mapping `sess.run([a, b])` -> `tf.get_default_session().run([a, b])`, I would use two lines and do `sess = get_session()` followed by `sess.run([a, b])`.

I'd also add these two functions to Edward's `__init__.py` so users can easily use them in their own scripts. That is, the line `from .util import set_seed` now becomes `from .util import get_session, set_seed, set_session`.

mariru:
Thanks for the feedback! I'll incorporate it.

There is a problem with the suggested `set_session()`. If the user specified a session using `tf.Session()` which is not interactive some of the syntax I have been using in this branch won't work (e.q `t.eval()`).

So one option is to work only with sessions which are not interactive or to allow only for interactive sessions to be passed into `set_session()` or even not incorporate this function at all.

Thoughts?

dustinvtran:
Hmm I guess there's no way to convert a Session object into a InteractiveSession object. In the `set_session(sess)` we can check if it's an InteractiveSession. If it is, then it will set it; it if isn't then we can throw an error and say the session must be an InteractiveSession?

mariru:
* I incorporated `get_session()` into `util.py` as suggested.
There is no need to make `_ED_SESSION` a global variable.

```
def get_session():
    """Get the default session; if not already defined, then
    the function will create an interactive session."""
    if tf.get_default_session() is None:
        _ED_SESSION = tf.InteractiveSession()
    _ED_SESSION = tf.get_default_session()
    return _ED_SESSION
```
* `set_session()` will be omitted. If a user already has an interactive session running we are able to access it in `ed.get_session()` using `tf.get_default_session()`. No need to set it.

dustinvtran:
I see. So `tf.get_default_session()` will already find that previously defined session if it's an interactive session but not if it's a typical session. For example,
```python
sess = tf.InteractiveSession()
tf.get_default_session() # returns the session

# exit python; re-enter python
sess = tf.Session()
tf.get_default_session() # returns None
```
Is there any way to error check if the user has already defined a session and that it's a non-interactive session? Ideally `get_session()` raises an error if the user has done so.

PR #118, opened 2016-06-05
Include models in setup.py

dcjones:
Version 1.0.6 fails to import because the `models` module isn't installed. This should fix that.


dustinvtran:
Thanks!

PR #120, opened 2016-06-14
added probit regression example

diengadji:
## Summary:

 1 - solves the NaN issue with hierarchical_logistic_regression example for real datasets
 2 - user can change hyper parameters prior_variance and sigma to avoid those NaN values
 3 - converges faster and to a better solution

## Intended Effect:

 provide another example

## How To Verify:

 run it

## Side Effects:

 None

## Documentation:

None

## Reviewer Suggestions:

@dustinvtran

### DIFF
--- a/examples/probit_regression.py
+++ b/examples/probit_regression.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python
+"""
+Hierarchical logistic regression using mean-field variational inference.
+
+Probability model:
+    Hierarchical logistic regression
+    Prior: Normal
+    Likelihood: Bernoulli-Logit
+Variational model
+    Likelihood: Mean-field Normal
+"""
+import edward as ed
+import tensorflow as tf
+import matplotlib.pyplot as plt
+import numpy as np
+from edward.util import get_dims
+from edward.models import Variational, Normal
+from edward.stats import bernoulli, norm
+
+class ProbitReg:
+    """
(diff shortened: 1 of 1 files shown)


dustinvtran:
Hi @diengadji, thanks for the PR. Does the hierarchical logistic regression not currently work? It seems to be working fine for me. Also is the only difference in this from the hierarchical logistic regression example the probit link over logit?

diengadji:
Hi @dustinvtran! the hierarchical logistic regression works when I use the build_toy_dataset method but when I use it on the crabs data from UCI repository for example I get NaN after a few iterations. The parametrization in terms of latent variables \phi that have variance sigma will help avoid those nans by tuning sigma...the only difference with your hierarchical regression is the probit link and the new parametrization. So I guess we can just modify the hierarchical logistic regression example to include those two modifications...

dustinvtran:
Thanks for the info! That would be okay although it wouldn't be hierarchical logistic regression any longer. And it'd be nice to get hierarchical logistic regression numerically stable. I haven't noticed any stability issues with the build toy data set, but if there are for the crabs data, can you send here a reproducible script that displays this behavior? 

Because logit and probit have very little difference, it should be easy for the user to work off just the hierarchical logistic regression example if they do want probit. (An example of something perhaps useful on the repo would be, e.g., hierarchical multinomial regression, as the data representation of one-hot or categorical is non-trivial and not "too" easy to extend.)

PR #136, opened 2016-06-28
initial version of the bbvi tutorial...to be updated

diengadji:
remaining to be done:
1 - plot elbo
2 - plot decision boundary
3 - add few lines on internals of one of the gradients


dustinvtran:
Hi @diengadji, great work! Here are some comments:
  + "score gradient" -> "score function gradient"
  + Can you elaborate more on that symbology for the reparameterization
  gradient? It can be hard for people to understand if they see that
  for the first time.
  + We now use `print(variational)` instead of
  `variational.print_params()`.
  + Can you add an emptyline between the print and the plot line in the last code
  snippet?
  + Your model needs fixing for the indices, e.g., you should use
  `np.reshape()` for `x` and not `self.reshape`; also the mapping()
  function's indices in its implementation look different from the
  current script which works. Is there a reason for that?

diengadji:
@dustinvtran thanks for your comments! I will integrate them to the updated version.

For the model, I had to modify the indices for the case D=2 since it was not running correctly...but that still needs to be fixed as you mentioned...

diengadji:
@dustinvtran : I updated the tutorial. Let me know if you have any further comments.

dustinvtran:
Looks great! My remaining comments are only comments on the writing.

> Black box variational inference is a Bayesian inference method that has been introduced by Paisley et al. and later by Ranganath et al.. 

replace with "Black box variational inference is a Bayesian inference method introduced by Ranganath et al. (2014), with related work by Paisley et al. (2012) and Mnih and Gregor (2014)."

> Adopting the Bayesian framework, you put a prior on z and aim to learn the posterior distribution

replace with "Given a model which is specified through a joint distribution $p(x, z)$, we aim to infer the posterior distribution".

> all equations

Can you add punctuation inside all equations where they are appropriate? In general, we go by the convention that [a mathematical text is, before everything else, a text](http://mathoverflow.net/questions/6675/periods-and-commas-in-mathematical-writing).

> equations with E_q

In general, I think the convention should be `\mathbb{E}_{q(z; lambda)} [ log p(x, z) - log q(z) ]`.  This features multiple things:
+ Use log - log instead of a log \frac{}. This makes the single line cleaner.
+ All expectation operators have brackets to organize the inputs to the operators.
+ All expectation operators are subscripted by a distribution `q(z; lambda)`, and not just a distribution `q` which can often be unclear what it means.
+ The parameters to a distribution are given with `;` instead of `|`. Only do parameters with distributions get `|`. This is an important distinction especially in hierarchical models with parameters to optimize vs parameters to infer a distribution.

> ELBO, KL

Can you use `\mathrm{ELBO}` and `\mathrm{KL`}?

> The reparametrization gradient when z=g(λ,ϵ) and ϵ∼p(ϵ) is:

I think you should elaborate more on those math phrases mean in words. It can be unclear to most people who don't understand the reparameterization trick and why it does anything useful. For example: "In a special class of models, we can also apply the reparameterization gradient. It samples from a base distribution `\epsilon \sim p(\epsilon)` and applies a transformation `z = z(\epsilon; \lambda)` such that the output is equivalent to having drawn from the original distribution. In this class, we can re-write the ELBO as".

> For a beta distribution these are the shape and scale parameters etc.

replace with "For a beta distribution these are the shape and scale parameters."

> Let's generate and visualize some toy data that we can illustrate our example with. Unfortunately Edward does not have yet data generators. This will be in subsequent versions. However we can generate data very easily as follows:

replace with "Let's generate and visualize some toy data that we can illustrate our example with."

> where we set the seed to 10 for reproducibility.

I'd remove this sentence altogether. (Also according to the code it doesn't actually set the seed to 10)

> dimentional

replace with "dimensional"

> We can simply sample w from the variational distribution and use that value to compute the score.

I wasn't sure what this meant. What's the score? It seems like from your code, `pred_function` is a function which returns the predicted class. Can you use something like "predicted class" instead of "score"?

> The second option is to compute the posterior predictive distribution.

Can you also add the posterior predictive distribution equation after that sentence, i.e.,
```
\begin{equation}
p(x_{\rm new} | x) = \int p(x_{\rm new} | z) p(z | x) dz
\end{equation}
```

dustinvtran:
An additional comment: In general, git commits should only occur after a reasonable number of changes are made, and a single message explains all those changes. I would avoid commiting as often as it can pollute the git commit history, particularly with the same message.

diengadji:
@dustinvtran @akucukelbir : Final version in. Thanks for your feedback and help.

dustinvtran:
Perfect. One last thing before merging: Can you squash all commits on this branch into a single commit? http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Afterwards, to push the changes, be sure to force push only onto this branch, e.g., `git push --force origin bbvi_tutorial`.

Alternatively, since it's only a single file you can make an alternative branch, add the file, make the commit, and then submit another PR and we'll close this one.

Issue #80, opened 2016-05-13
add distribution objects


dustinvtran:
Any distribution available in `edward.stats` should have a corresponding distribution object that is also available in `edward.models`.

Discrete
+ [x] Bernoulli
+ [ ] Binomial
+ [x] Multinomial (assume single trial, i.e., `n=1` when sampling and calculating density)
+ [ ] Poisson
+ [ ] Geometric
+ [ ] Negative Binomial
+ [x] Point mass

Continuous
+ [ ] Uniform
+ [ ] Multivariate Normal
+ [ ] Student-t (with mean and scale parameters)
+ [ ] Gamma
+ [x] Inverse Gamma
+ [ ] Exponential
+ [x] Beta
+ [x] Dirichlet
+ [ ] Log-Normal
+ [ ] Chi-squared

See also #77.

altosaar:
Should we stick to tf.contrib.distributions and submit PRs there?

dustinvtran:
(looping in @heywhoah. I'm also planning to discuss this next week in the Bayesflow meeting.)

Yeah I was thinking about this recently. The distribution objects in our library and the distribution objects in `tf.contrib.distributions` are very similar. There are a few things to address.

1. The two distribution objects have slightly different desired behavior. For example, the inputs/outputs to our log density method is different. In another branch, we are adding arithmetic so that we can treat the distribution objects as if they were any other tensor, and that we can also bind them to realizations. This will make our objects inbetween distribution objects in `tf.contrib.distributions` and distribution tensors in `tf.contrib.bayesflow.scg`.

  __Solution.__ We can use `tf.contrib.distributions` as the base, and have our distribution objects be a wrapper around it. This enables higher-level methods  without having to reinvent the wheel with regards to primitive operations. It is analogous to how `tf.contrib.bayesflow.scg` builds on top of `tf.contrib.distributions`. 

  We will follow TensorFlow's standards regarding input/output behavior, and also style guidelines. This will make it easier to submit pull requests as we develop primitive operations that are not available in `tf.contrib.distributions` that we need. (We have many of these.)
2. Experimental features in TensorFlow. We only want to have a stable version of TensorFlow as a dependency, and not require features from `tf.contrib` which is bound to change.

  __Solution.__ Perhaps we should have a static version of `tf.contrib.distributions` natively in Edward. That is, we copy and paste the whole `tf.contrib.distributions` as a subdirectory in Edward?

dustinvtran:
[...]

PR #138, opened 2016-06-29
Feature/shapes

dustinvtran:
#### Summary:

+ issues: makes progress for #7; makes progress for #27; helps refactor for #80 in moving to `tf.contrib.distributions`

All distribution objects, i.e., any class inheriting from `Distribution` in `edward.models`, now takes as input a shape and optionally, parameter arguments. The shape denotes the shape of its random variable. For example:
```python

# 5 x 2 matrix of random variables
x = Normal([5, 2])
x.sample(size=10)
```
The output is a set of samples with shape `(10, 5, 2)`, where size is the outer dimension.

This generalizes the previously cryptic `num_factors` argument and enables tensors of random variables. Note also that as a result, all sample methods—whether in be in the container object `Variational()`, distribution objects `Distribution()`, or `edward.stats` sampling `rvs()`—will draw samples where the size is the outer dimension. For example:
```python
from edward.stats import bernoulli

x = bernoulli.rvs(p=0.5, size=1)
print(x.shape)
## (1,)
x = bernoulli.rvs(p=np.array([0.5]), size=1)
print(x.shape)
## (1, 1)
x = bernoulli.rvs(p=np.array([0.5, 0.2]), size=3)
print(x.shape)
## (3, 2)
```
This goes against SciPy behavior (which is inconsistent anyways): `size=1` typically results in the same output dimension as the parameter rather than `size x shape(parameter)`. Our standard adheres to the behavior in the `tf.contrib.distributions` library.

The container object `Variational()` now works with a list of tensors if there is more than one layer, and a single tensor if only one layer. This arises in the input for `variational.log_prob(xs)` as well as the output for `variational.sample(size)`. Carrying around a list of tensors makes packing and unpacking of latent variables much easier to work with; see the mixture of Gaussians example. Carrying around a tensor if only one layer is convenient for probability models that deal with only one set of latent variables; it's easier to work directly with `zs` rather than do `zs[0]`; see the Beta-Bernoulli or hierarchical linear model examples.

There is a nuance worth mentioning why there's a difference in the `log_prob(xs, zs)` methods of `mixture_gaussian.py` compared to `mixture_gaussian_map.py`. The former uses three sets of variational distributions; the latter uses one (a point mass). This means the former takes `zs` as a list of 3 tensors, and the latter takes `zs` as a single tensor. While this isn't satisfactory (the probability model's method should not rely on the variational model downstream), this makes the difference which already currently exists more transparent.

Finally, I also simplified the `log_prob()` method. All current inference algorithms use a more holistic `variational.log_prob()` rather than explicitly summing over `variational.log_prob_idx()`. We can deal with the latter in future special cases when we do in fact want to take advantage of individual factors.

Extensions
  + __Matrixvariate random variables, tensorvariate, etc.__ We currently use a flag in the distribution obejcts to check if `is_multivariate`, which is important for indexing in the methods. We can later generalize the boolean to some integer value, e.g., 0 if univariate, 1 if vectorvariate, 2 if matrixvariate, etc.
  + __Smart arguments for the distribution object.__ For example, we may want to do `Normal(3, loc=1, scale=scale)`, where `scale` is `(3, 5)` and the 1 in `loc` automates to `(3, 5)`.
  + __Shape of samples when size=1.__ In a previous series of commits, I used the output `shape` if size = 1 and `size x shape` if size > 1. We can bring this back later on when it may be useful. For example, `tf.contrib.bayesflow.scg` allow this through the `SampleResampleValue` data type.
  + __Higher-order shapes.__ While the construction of the distribution and container objects is general, we are limited to at most matrix random variables because the primitives it calls upon in `edward.stats` are not guaranteed to work for 3-d or higher tensors. This includes: `rvs()`, `logpmf()`/`logpdf()`, `to_simplex()`; and `entropy()` is not guaranteed to work for 2-d or higher tensors if the distribution is univariate.

#### Intended Effect:

See above.

#### How to Verify:

Run all examples and tests. They still work.

We need additional unit tests for three things:
+ container object
+ distribution objects
+ [x] `rvs()` in `edward.stats`

#### Side Effects:

None.

#### Documentation:

I added and clarified documentation to all relevant methods.

#### Reviewer Suggestions:

N/A


dustinvtran:
Thanks @akucukelbir for reviewing. Also looping in @heywhoah because it's one of the first steps toward the modeling language.

edward/inferences.py
```diff
@@ -199,12 +199,10 @@ def build_score_loss(self):
         ELBO = E_{q(z; lambda)} [ log p(x, z) - log q(z; lambda) ]
         """
         x = self.data.sample(self.n_data)
-        z, self.samples = self.variational.sample(self.n_minibatch)
-
-        q_log_prob = tf.zeros([self.n_minibatch], dtype=tf.float32)
-        for i in range(self.variational.num_factors):
-            q_log_prob += self.variational.log_prob_i(i, tf.stop_gradient(z))
+        self.samples = self.variational.sample(self.n_minibatch)
+        z = self.samples
```
akucukelbir: does `self.samples` always/only contain samples from the variational model?

if so, consider renaming it `self.z_samples` and avoid statements like `z = self.samples`.

edward/inferences.py
```diff
@@ -376,7 +368,7 @@ def __init__(self, model, data=Data(), params=None):
 
     def build_loss(self):
         x = self.data.sample(self.n_data)
-        z, _ = self.variational.sample()
+        z = self.variational.sample()
         self.loss = tf.squeeze(self.model.log_prob(x, z))
```
akucukelbir: why doesn't `MAP` store variational samples into `self.samples`?

mariru: Because for `MAP` `self.variational.sample` simply returns the current value of the latent variables. No samples are needed elsewhere so they aren't stored.

edward/models/distributions.py
```diff
@@ -70,45 +73,80 @@ def sample(self, size=1):
...
             if layer.sample_tensor:
                 samples += [layer.sample(size)]
             else:
-                samples += [tf.placeholder(tf.float32, (size, layer.num_vars))]
+                samples += [tf.placeholder(tf.float32, (size, ) + layer.shape)]
 
-        return tf.concat(1, samples), samples
+        if len(samples) == 1:
+            samples = samples[0]
+
```
akucukelbir: what does this squeezing buy us? does it only affect us when we draw a single sample?

edward/inferences.py
dustinvtran: Yep. Something like `self.z_samples` makes sense.

edward/inferences.py
akucukelbir: okay thank you.

minor suggestion: rename `z` to something like `point_estimate_z`?

edward/models/distributions.py
```diff
@@ -70,45 +73,80 @@ def sample(self, size=1):
...
+        return samples
 
     def np_dict(self, samples):
         """
         Form dictionary to feed any placeholders with np.array
         samples.
+
+        Parameters
+        ----------
+        samples : list or tf.Tensor
```
akucukelbir: seems like we also accept `np.array-like` ? see line 103 below

edward/models/distributions.py
dustinvtran: It's useful to distinguish between the container of distribution objects and the shape of their samples. The above is unrelated to the number of drawn samples. It's relates to whether the container only holds one layer or it holds multiple layers.

edward/models/distributions.py
```diff
@@ -70,45 +73,80 @@ def sample(self, size=1):
...
         """
+        if not isinstance(samples, list):
+            samples = [samples]
+
+        size = get_dims(samples[0])[0]
```
akucukelbir: this is a bit cryptic. `size` = the number of "samples" right? could we name it as such?

edward/models/distributions.py
```diff
@@ -70,45 +73,80 @@ def sample(self, size=1):
...
+            return self.layers[0].log_prob(xs)
 
-            i = i - layer.num_factors
-            start = final
+        if isinstance(xs[0], tf.Tensor):
+            shape = get_dims(xs[0])
+            rank = len(shape)
+        else: # NumPy array
+            shape = xs[0].shape
+            rank = len(shape)
```
akucukelbir: what is `rank`? do we need it anymore?

edward/models/distributions.py
```diff
@@ -119,19 +157,31 @@ def entropy(self):
...
+    where shape is the shape of x and params[idx] denote the
+    parameters of each random variable x[idx].
 
     Parameters
     ----------
-    num_factors : int
-        Number of factors.
+    shape : int, list, or tuple, optional
+        Shape of random variable(s). For multivariate distributions,
+        the outermost dimension denotes the multivariate dimension.
```
akucukelbir: "outermost" confusion

akucukelbir: also, should mention connection of last dimension and how it relates to `is_multivariate`

edward/models/distributions.py
```diff
@@ -182,33 +232,106 @@ def sample(self, size=1):
...
+            else:
+                for i in range(self.shape[0]):
+                    for j in range(self.shape[1]):
+                        for k in range(self.shape[2]):
+                            idx = (i, j, k, )
+                            log_prob += self.log_prob_idx(idx, xs)
+
+        else: # len(self.shape) >= 4
+            # There should be a generic recursive solution.
+            raise NotImplementedError()
```
akucukelbir: is the issue the input `idx` into `self.log_prob_idx`? if so would this work?

http://stackoverflow.com/questions/22473053/itertools-equivalent-of-nested-loop-for-x-in-xs-for-y-in-ys

edward/models/distributions.py
```diff
@@ -241,39 +364,33 @@ def __str__(self):
...
 
-    def log_prob_i(self, i, xs):
-        """log p(x_i | params)"""
-        if i >= self.num_factors:
-            raise IndexError()
-
-        return bernoulli.logpmf(xs[:, i], self.p[i])
+    def log_prob_idx(self, idx, xs):
+        """log p(xs[:, idx] | params[idx])"""
+        full_idx = (slice(0, None), ) + idx
```
akucukelbir: `(slice(0, None), )` ? perhaps this warrants a short inline comment?

edward/models/distributions.py
```diff
@@ -385,55 +489,44 @@ def sample(self, size=1):
...
-            x = tf.sigmoid(eq + pi_unconst)
-            pil = tf.concat(1, [x, tf.ones([self.num_factors, 1])])
-            piu = tf.concat(1, [tf.ones([self.num_factors, 1]), 1.0 - x])
-            # cumulative product along 1st axis
-            S = tf.pack([cumprod(piu_x) for piu_x in tf.unpack(piu)])
-            pi = S * pil
+            real_shape = self.shape[:-1]
+            K_minus_one = self.shape[-1] - 1
+            pi_unconst = tf.Variable(tf.random_normal([real_shape + (K_minus_one, )]))
+            pi = to_simplex(pi_unconst)
```
akucukelbir: nice!

edward/stats/distributions.py
```diff
@@ -9,68 +9,139 @@ class Distribution:
...
+
+        params : float, np.array, or tf.Tensor
+
         Returns
         -------
         np.ndarray
-            size-dimensional vector; scalar if size=1
+            np.array of dimension (size x shape), where shape is the
+            shape of its parameter argument. For multivariate
+            distributions, shape corresponds to one of the parameter
```
akucukelbir: ``shape`` corresponds to the dimension of the multivariate random variable, e.g., the dimension of ``alpha`` in a ``Dirichlet``, the dimension of ``p`` in a ``Multinomial``, or the dimension of ``mean`` in ``Multivariate_Normal``

edward/stats/distributions.py
```diff
@@ -9,68 +9,139 @@ class Distribution:
...
-            If univariate distribution, can be a scalar, vector, or matrix.
-            If multivariate distribution, can be a vector or matrix.
+        x : float, np.array, or tf.Tensor
+            If univariate distribution, can be a scalar or tensor.
+            If multivariate distribution, can be a tensor; the outer
+            dimension carries the multivariate dimension.
 
-        params : np.array or tf.Tensor
+        params : float, np.array, or tf.Tensor
             scalar unless documented otherwise
```
akucukelbir: which `params`?

edward/stats/distributions.py
```diff
@@ -9,68 +9,139 @@ class Distribution:
...
+        >>> x = bernoulli.rvs(p=0.5, size=1)
+        >>> print(x.shape)
+        (1,)
+        >>> x = bernoulli.rvs(p=np.array([0.5]), size=1)
+        >>> print(x.shape)
+        (1, 1)
+        >>> x = bernoulli.rvs(p=np.array([0.5, 0.2]), size=3)
+        >>> print(x.shape)
+        (3, 2)
+        """
```
akucukelbir: i really like this. i'll endeavor to include this into the `docstrings` down the road

edward/models/distributions.py
akucukelbir: or how about http://docs.scipy.org/doc/numpy/reference/generated/numpy.nditer.html

edward/stats/distributions.py
```diff
@@ -9,68 +9,139 @@ class Distribution:
...
+            p = np.asarray(p)
+        if len(p.shape) == 0:
+            return stats.bernoulli.rvs(p, size=size)
+
+        x = []
+        for pidx in np.nditer(p):
+            x += [stats.bernoulli.rvs(pidx, size=size)]
+
+        # Note this doesn't work for multi-dimensional sizes.
+        x = np.asarray(x).transpose()
```
akucukelbir: as in when `len(size)>=3` ?

edward/stats/distributions.py
```diff
@@ -83,7 +154,20 @@ def entropy(self, p):
...
+        if not isinstance(a, np.ndarray):
+            a = np.asarray(a)
+        if not isinstance(b, np.ndarray):
+            b = np.asarray(b)
+        if len(a.shape) == 0:
+            return stats.beta.rvs(a, b, size=size)
+
+        x = []
+        for aidx,bidx in zip(np.nditer(a), np.nditer(b)):
+            x += [stats.beta.rvs(aidx, bidx, size=size)]
```
akucukelbir: space between `aidx, bidx`

edward/stats/distributions.py
```diff
@@ -189,7 +319,20 @@ def entropy(self, scale=1):
...
+        if not isinstance(a, np.ndarray):
+            a = np.asarray(a)
+        if not isinstance(scale, np.ndarray):
+            scale = np.asarray(scale)
+        if len(a.shape) == 0:
+            return stats.gamma.rvs(a, scale=scale, size=size)
+
+        x = []
+        for aidx,scaleidx in zip(np.nditer(a), np.nditer(scale)):
+            x += [stats.gamma.rvs(aidx, scale=scaleidx, size=size)]
```
akucukelbir: space `aidx, scaleidx`

edward/stats/distributions.py
```diff
@@ -218,7 +372,20 @@ def entropy(self, p):
...
+        if not isinstance(a, np.ndarray):
+            a = np.asarray(a)
+        if not isinstance(scale, np.ndarray):
+            scale = np.asarray(scale)
+        if len(a.shape) == 0:
+            return stats.invgamma.rvs(a, scale=scale, size=size)
+
+        x = []
+        for aidx,scaleidx in zip(np.nditer(a), np.nditer(scale)):
+            x += [stats.invgamma.rvs(aidx, scale=scaleidx, size=size)]
```
akucukelbir: ditto

edward/stats/distributions.py
```diff
@@ -254,7 +432,21 @@ def entropy(self, s):
...
+            # np.random.multinomial defaults to (size x p.shape)
+            return np.random.multinomial(n, p, size=size)
+
+        if not isinstance(n, np.ndarray):
+            n = np.asarray(n)
+
+        x = []
+        # This doesn't work for non-matrix parameters.
+        for nidx,prow in zip(n, p):
+            x += [np.random.multinomial(nidx, prow, size=size)]
```
akucukelbir: ditto

edward/stats/distributions.py
```diff
@@ -311,7 +503,23 @@ def entropy(self, n, p):
...
+            # stats.multivariate_normal.rvs returns (size x shape) if
+            # size > 1, and shape if size == 1.
+            if size == 1:
+                return np.expand_dims(x, axis=0)
+            else:
+                return x
+
+        x = []
+        # This doesn't work for non-matrix parameters.
+        for meanrow,covmat in zip(mean, cov):
```
akucukelbir: ditto

edward/stats/distributions.py
```diff
@@ -443,7 +688,22 @@ def entropy(self, mu):
...
     def rvs(self, df, loc=0, scale=1, size=1):
-        return stats.t.rvs(df, loc=loc, scale=scale, size=size)
+        if not isinstance(df, np.ndarray):
+            df = np.asarray(df)
+        if not isinstance(loc, np.ndarray):
+            loc = np.asarray(loc)
+        if not isinstance(scale, np.ndarray):
+            scale = np.asarray(scale)
+        if len(df.shape) == 0:
+            return stats.t.rvs(df, loc=loc, scale=scale, size=size)
```
akucukelbir: what if `len(df.shape)==0` but `len(loc.shape)!=0` ?

akucukelbir:
@dustinvtran okay. i reviewed the main files that changed. i didn't look closely at the examples or unit tests. overall, fantastic fantastic work.

edward/inferences.py
dustinvtran: I prefer `z` because it follows semantically from what is happening, i.e., we are sampling from `variational` to get a 1 x shape(z) output, where `z` is a point mass family.

edward/inferences.py
dustinvtran: I'll use `self.zs` because we use `zs` in `log_prob(xs, zs)` methods in the model. This makes it more consistent. We can think about whether it should be more verbose that it represents samples explicitly in the variable name or in the docs later on.

dustinvtran: I'm also keeping the statement `z = self.zs` so that the rest of the code is clean in applying functions to `x` and `z`. We can think about whether to not do this statement later.

edward/models/distributions.py
dustinvtran: `np_dict` takes as input a list of tf.Tensors or a tf.Tensor. This is directly the output of the `sample()` method in the container. It is so that it can bind any TensorFlow placeholders with realizations (`np.arrays`).

edward/models/distributions.py
dustinvtran: `size` is used as the number of samples in other contexts, e.g., SciPy stats sampling. @mariru is planning to change this naming so it's consistent with `n_minibatch`.

edward/models/distributions.py
dustinvtran: Great point. I used it in previous commits and don't need it anymore.

edward/models/distributions.py
dustinvtran: Given a tuple (`self.shape`), we'd like to iterate over all tuples where each element in the tuple is less than the corresponding element in `self.shape`. This forms all the indices we sum over in `log_prob = log p(xs | params) = sum_{idx in shape} log p(xs[idx] | params[idx])`.

Your first suggestion lets us collapse the nested for loops into one, although it adds an additional library dependency (`itertools`). In your second suggestion, I think we ultimately want something like `multi_index` in `np.nditer()`. But I haven't found out how to do it precisely.

edward/inferences.py
akucukelbir: okay let's pass for now.

i made an issue: https://github.com/blei-lab/edward/issues/139

edward/inferences.py
akucukelbir: very well. carry on.

edward/models/distributions.py
akucukelbir: yes, but why can't we do `sample[0]` outside of this function? wouldn't that be more transparent?

edward/models/distributions.py
akucukelbir: i see. 

type checking would make this clearer. but we can leave that for later. i think i'll eventually want to start a refactor project where we figure out a scheme of raising and catching type exceptions

edward/models/distributions.py
akucukelbir: very well. carry on.

edward/models/distributions.py
akucukelbir: ok. i made an issue. https://github.com/blei-lab/edward/issues/140

edward/models/distributions.py
dustinvtran: we could but it ruins the current interface for most of the example scripts. they would have to work on `zs[0]` rather than `zs`.

edward/stats/distributions.py
dustinvtran: when size is a list and not just an integer

edward/stats/distributions.py
dustinvtran: i added doc to explain this. basically, the parameters must have consistent shapes/types

Issue #123, opened 2016-06-20
option for sparse data representation in data class


mariru:

dawenl:
I am very interested in this as well -- and it's highly relevant to what I mostly work with. Do you have any plan on what to do? My concern is that among all the languages/frameworks that edward supports, at least tensorflow doesn't have a very good sparse support, which could be a problem.

mariru:
I had something like this in mind:
Store the existing data as a one dimensional tensorflow tensor and also store the indices that relate this sparse representation to the dense representation which is not stored explicitly.

Then, when `data.sample()` is called it returns both a subsample of the data as well as the indices this values correspond to.

dustinvtran:
So at the moment data is only used in the model (and mapping if the user specifies an inverse mapping for `q(z | x)` that's non-trivial). Is there a reason sparse data is not supported if the user defines both the data object and the method (`model.log_prob()`) which interfaces with the data object? Or do you mean it's non-trivial how to do data subsampling?

dawenl:
I think the solution @mariru proposes will work with subsampling, but my concern is that it will not work when there is a need for sparse operation (e.g. matrix dot product between the sparse data and parameter) which should be pretty common.

dawenl:
Just found out about [tf.SparseTensor](https://www.tensorflow.org/versions/r0.9/api_docs/python/sparse_ops.html), which seems a good solution to this. Anyone has any experience with it? If not, I can take a stab at it.

dustinvtran:
Nope, and I also recommend you can take a stab at it :)

dustinvtran:
I still think this is fine because only the user interacts explicitly with the data when they work with the `log_prob()` method in their defined model. So I'm not sure where I see the precise difficulty but maybe this will be clearer to me as you do stuff.

I do see a need for sparse tensors when dealing with sparse Distribution objects though. This will become clear after I merge #138, which lets distribution objects have arbitrary shape to represent arbitrary-sized tensors of random variables.

dustinvtran:
closing as we no longer have a data class (nor do we hand-implement any distribution methods)

Issue #26, opened 2016-02-28
streaming data


dustinvtran:
Implementation-wise, this is the setting where we can't store all of the data inside the `Data` class. What's the best way to couple this problem when streaming is integrated into the iterations of `inference.run()`?

mariru:
I want to revive this issue. 
There are examples like the [mixture density network](https://github.com/blei-lab/edward/blob/master/examples/tf_mixture_density_network_demo.py) where data is a `tf.placeholder` in the model and then fed during inference using `feed_dict`.
Really what is missing is a general method to subsample data during inference for the `feed_dict` similar to the functionality of `mnist.train.next_batch()` in the tensorflow mnist tutorial.

mariru:
@dustinvtran @dawenl 
A general solution could do something similar as this (the code snippet is part of an example I am currently working on).
As Dawen suggested in #25  I am using yield.

```
# DATA
data = {'x': x, y: l}
n_minibatch = 10

def splice(x,i,j,l):
    if j>=l:
        return np.hstack((x[i:],splice(x,0,j-l,l)))
    else:
        return x[i:j]
        
def batch_generator(x,n_minibatch=10):
    j = 0
    l = len(x)
    while True:
        i = j%l
        j = i+n_minibatch
        yield splice(x,i,j,l)
        
train_batches = batch_generator(np.loadtxt('../../dat/klijn/train_idx.txt').astype(int), n_minibatch)
test_batches = batch_generator(np.loadtxt('../../dat/klijn/test_idx.txt').astype(int), n_minibatch)

# INFERENCE
for i in range(n_iter):
    train_idx = train_batches.next()
    test_idx = test_batches.next()
    _, train_loss[i] = sess.run([inference.train, inference.loss],
                              feed_dict={x: features[train_idx,:], l: labels[train_idx,:]})
    test_loss[i] = sess.run(inference.loss,
                              feed_dict={x: features[train_idx,:], l: labels[train_idx,:]})
    print("Train Loss: {:0.3f}, Test Loss: {:0.3f}".format(train_loss[i],
                                                         test_loss[i]))
```

dustinvtran:
For streaming from files, I recommend using TensorFlow's [file readers](https://www.tensorflow.org/versions/master/how_tos/reading_data/index.html#reading-from-files). Basically, they take a file as input and output a data tensor. The data values in the tensor will change at each session run (according to settings in the file reader). Edward's inference can operate on these data tensors as if they were just any other data. 

[...]

Issue #25, opened 2016-02-28
data subsampling for more complicated data structures


dustinvtran:
How do we subsample data for matrices, ragged arrays (e.g., time series), tensors, etc.?

dustinvtran:
See prescribed solution [here](https://github.com/blei-lab/edward/issues/26#issuecomment-249470497).

PR #156, opened 2016-07-05
Feature/data

### LINKED ISSUE #56: unifying data specification
dave brought up the idea of abstracting the data specification: the users will have one way to specify the data and in the backend the data provided will be morphed into a stan data model, or a tensorflow data model etc...

### LINKED ISSUE #95: update documentation on the different ways you can specify data
#90 added a lot of new ways to specify data. We should write this up because it can be confusing without understanding the different options and when/how to use them.

### LINKED ISSUE #112: predict method for model in evaluation tasks
The current inputs and outputs of `model.predict()` are rather unnatural. In principle, it shouldn't have access to the test output, and only the test input. For this to apply, the Data class must have a notion that distinguishes between output data and input data.

dustinvtran:
#### Summary:

+ issues: fixes #56; fixes #95; fixes #112; makes progress for #27; related to #25; related to #26; related to #123

This feature provides one data abstraction to rule all them all. It is the simplest abstraction satisfying all desired features: a Python dictionary.

Data as a Python dictionary bodes nicely with TensorFlow's `feed_dict` concept: in the same way `feed_dict` is a dictionary binding TensorFlow placeholders to NumPy arrays, `data` is a dictionary binding random variables to realizations. This connection will be exploited even more as we implement the modeling language.

How do the different languages interact with the data dictionary?
+ TensorFlow: The dictionary carries whatever keys and values the user accesses in `log_prob()` (or in the other user-defined methods). Key is a string. Value is a NumPy array or TensorFlow placeholder.
```python
class BetaBernoulli:
    def log_prob(self, xs, zs):
        log_prior = beta.logpdf(zs, a=1.0, b=1.0)
        log_lik = tf.pack([tf.reduce_sum(bernoulli.logpmf(xs['x'], z))
                           for z in tf.unpack(zs)])
        return log_lik + log_prior

model = BetaBernoulli()
data = {'x': np.array([0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}
```
+ Python: The dictionary carries whatever keys and values the user accesses in `log_prob()` (or in the other user-defined methods). Key is a string. Value is a NumPy array or TensorFlow placeholder.
```python
class BetaBernoulli(PythonModel):
    def _py_log_prob(self, xs, zs):
        n_minibatch = zs.shape[0]
        lp = np.zeros(n_minibatch, dtype=np.float32)
        for b in range(n_minibatch):
            lp[b] = beta.logpdf(zs[b, :], a=1.0, b=1.0)
            for n in range(len(xs['x'])):
                lp[b] += bernoulli.logpmf(xs['x'][n], p=zs[b, :])

        return lp

model = BetaBernoulli()
data = {'x': np.array([0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}
```
+ PyMC3: The dictionary binds Theano shared variables, which are used to mark the observed PyMC3 random variables, to their realizations. Key is a Theano shared variable. Value is a NumPy array or TensorFlow placeholder.
```python
x_obs = theano.shared(np.zeros(1))
with pm.Model() as pm_model:
    beta = pm.Beta('beta', 1, 1, transform=None)
    x = pm.Bernoulli('x', beta, observed=x_obs)

model = PyMC3Model(pm_model)
data = {x_obs: np.array([0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}
```
+ Stan: It's the usual dictionary according to the Stan program's data block. Key is a string. Value is whatever type is used for the data block.
```python
model_code = """
    data {
      int<lower=0> N;
      int<lower=0,upper=1> y[N];
    }
    parameters {
      real<lower=0,upper=1> theta;
    }
    model {
      theta ~ beta(1.0, 1.0);
      for (n in 1:N)
        y[n] ~ bernoulli(theta);
    }
"""
model = ed.StanModel(model_code=model_code)
data = {'N': 10, 'y': [0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}
```
How do we use the data during training? Let's detail the three use cases.
+ Initialize training with full data. Loop over all data per iteration. (supported for all languages)

  Pass in the data via `inference = ed.MFVI(model, variational, data)`, then call `inference.run()`. See `examples/beta_bernoulli_tf.py` as an example.
+ Initialize training with full data. Loop over a batch per iteration. (scale inference in terms of computational complexity; supported for all but Stan)

  Pass in the data via `inference = ed.MFVI(model, variational, data)`, then call `inference.run(n_data=5)`. By default, we will subsample by slicing along the first dimension of every data structure in the data dictionary. See `examples/mixture_gaussian.py` as an example. 
+ Initialize training with no data. Manually pass in a batch per iteration. (scale inference in terms of computational complexity and memory complexity; supported for all but Stan)
  
  Define your data dictionary by using `tf.placeholder()`'s. Pass in the data via `inference = ed.MFVI(model, variational, data)`. Initialize via `inference.initialize()`. Then in a loop run `sess.run(inference.train, feed_dict={...})` where in the `feed_dict` you pass in the values for the `tf.placeholder()`'s. See `examples/mixture_density_network.py` as an example.

What kind of black magic is done internally to get this to work? All methods—model methods, inference methods, criticism methods—work with the data dictionary. For inference, we internally use a data abstraction which you can think of now as a data generator. This data generator simply outputs batches of data during training. The typical Edward user will not have to learn about it (the typical Edward inference researcher would).

__Extensions__
+ How do we deal with the different ways to [read data in TensorFlow](https://www.tensorflow.org/versions/r0.9/how_tos/reading_data/index.html#preloaded-data)?
    + Feeding. Solved with this pull request.
    + Reading from files. To be solved in another pull request.
    + Preloaded data. We don't aim to support this case because we don't see a particular use case for it. Just use a dictionary of NumPy arrays and follow the feeding approach instead.
+ More advanced and/or custom data generators. For convenient batch training when your data fits in memory, and using `yield`/iterator classes to avoid storing beyond that batch of data.
+ In general, there should be a scale factor due to data subsampling so that
    `log_lik \approx N / n_data * ( mini-batch log_lik )`
    for every type of log-likelihood. For now it is ignored as it has been before.

(This pull request has a long description. It's partially because I'm excited about this. It's partially because various pieces of the description will be in a tutorial explaining how to use data in Edward.)

#### Intended Effect:

See above.

#### How to Verify:

Run all examples and tests. They have been modified to work with data dictionaries.

#### Side Effects:

None.

#### Documentation:

I added and clarified documentation to all relevant methods.

#### Reviewer Suggestions:

N/A


dustinvtran:
Looping in @heywhoah because it relates to the modeling language

mariru:
Hey Dustin, this is great!

I ran all examples and they run without issue.

Here is a concern:
In the `Inference` class  we run `build_loss()` in `initialize()`. This is the only time we ever call.
``` 
x = self.data.next(self.n_data)
self.zs = self.variational.sample(self.n_minibatch)
```

While because of underlying tensorflow functionality `variational.sample()` will give us a new sample whenever the computational graph is evaluated, `data.next()`  is called only once. This means that the loss is always evaluated with the same batch of data (the one we generated in `initialize()`).

dustinvtran:
@mariru: Oh yes that is a grave error! I pushed a working solution. The internal implementation leaves much to be desired though. There are a bunch of additional functions floating around and I'm not sure of the best way to consolidate them (`self.variational.np_dict`, `self.data.make_placeholders`, `self.data.next` and their unintuitive methods and where they are placed). Will think about this a bit more.

mariru:
@dustinvtran:

I am still trying to understand your fix.
How about renaming the keys of the dictionaries as follows
`Placeholder` -> `dat_placeholder`
'Placeholder_1` -> `var_placeholder`

Why do we need to add the samples from the variational parameters to `feed_dict`?

dustinvtran:
@mariru: Okay, I finally have a clean implementation! Minor changes to the pull request description at the top. Here are the internal details:

+ All of inference now has two data-related members: `self.xs` and `self.data`. See their documentation in the code. `self.xs` is used to build the computational graph with TensorFlow placeholders; `self.data` is used to feed the placeholders at runtime.
+ The `DataGenerator` class is only a data generator for NumPy arrays. It takes nothing else. Its only method is `next()`, which returns slices along the first dimension of the NumPy array.

Let me know your thoughts.

mariru:
@dustinvtran this is great!

However `example/beta_bernoulli_stan.py` doesn't work because integers/lists in the data dictionary do not have a `shape()` attribute required for building placeholders for dictionary `self.data` in `inferences.py`.

Casting them into numpy arrays as follows
```
data = {'N': np.array([10]), 'y': np.array([0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}
```
doesn't work either. Stan's log probability function requires integers/floats/lists.

Another question is how can the user specify the minibatch size?

dustinvtran:
ill do a fix that fixes the data passed in for stan models then, as stan models dont allow data subsampling anyways.

re: minibatch size. its the same as usual: n_data

dustinvtran:
@mariru: You can take a look again.

PR #160, opened 2016-07-08
Feature/py func sample

dustinvtran:
This pull request comes at the advice of the Bayesflow folks. (also see https://github.com/tensorflow/tensorflow/issues/2423)

The sample method in distribution objects now always return a `tf.Tensor`. The internal implementation of enabling NumPy/SciPy samples has been modified. Rather than creating a placeholder to fill in NumPy-based samples later, we simply modify the `sample()` method to wrap around a Python function and call `tf.py_func`. See, e.g., Bernoulli, Beta, Dirichlet, Multinomial, and compare it to Normal and PointMass. 

This tightens up the internal code and makes it much easier to use `tf.contrib.distributions`-based sampling once they are available. (To do so, we simply rewrite the `sample()` method in the distribution object.)

Unit tests are added.

### Reviewer Suggestions:
@mariru: This addresses your question in #156 about what `variational.np_dict` does. The answer is that we don't actually need it. (hence this pull request!)

@akucukelbir: Following our chat today about SciPy based sampling and `tf.py_func`.

Either of you feel free to review this and merge it when you're done (or wait if you have particular comments).


akucukelbir:
@dustinvtran beautiful. love it. merge at will.

PR #161, opened 2016-07-08
remove need for sample_noise() and reparam() methods

### LINKED ISSUE #145: consider renaming sample_noise
`sample_noise` is unintuitive and not descriptive to people who haven't read the key papers.

consider replacing `sample_noise` with `sample_standard` or something that indicates that we are sampling from a fixed / standard / basic random variable in this function.

dustinvtran:
+ fix #145 (by no longer needing it!)

This pull request comes at the advice of the Bayesflow folks. (also see https://github.com/tensorflow/tensorflow/issues/2423)

Rather than explicitly defining a `sample_noise()` method and `reparam()` method in distribution objects, any reparameterizable distributions simply do the noise generation and reparameterization in `sample()`. This simplifies the internal code and is more closely aligned with how `tf.contrib.distributions` works; see, e.g., [`tf.contrib.distributions.Normal`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/normal.py#L304-L334).

### Reviewer Suggestions:

Anyone feel free to review and merge this.

### DIFF
--- a/edward/inferences.py
+++ b/edward/inferences.py
@@ -228,7 +228,7 @@ class MFVI(VariationalInference):
             gradient estimator. Otherwise default is to use the
             reparameterization gradient if available.
         """
-        if score is None and self.variational.is_reparam:
+        if score is None and self.variational.is_reparameterized:
             self.score = False
         else:
             self.score = True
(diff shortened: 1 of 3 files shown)


mariru:
Nice fix.
I checked the code and ran the examples.

PR #162, opened 2016-07-10
Cleanup/var names

mariru:
## Summary:
fixes the following conventions for variable naming:

number of samples from (variational) distribution:
`n_samples` 
number of data samples:
`n_minibatch`
number of variables:
`n_vars`  (used to be `num_vars`)
number of parameters:
`n_params` (used to be `num_params`)


edward/criticisms.py
```diff
@@ -28,6 +28,7 @@ def evaluate(metrics, model, variational, data, y_true=None):
         according to the Stan program's data block.
     y_true : np.ndarray or tf.Tensor
         True values to compare to in supervised learning tasks.
+    n_samples : number of posterior samples
```
dustinvtran: The first line in the doc string would be the type, e.g., `n_samples : int, optional` followed by `Number of posterior samples.`

edward/data.py
```diff
@@ -34,20 +34,20 @@ def __init__(self, data):
         self.N = data.shape[0]
         self.counter = 0
 
-    def next(self, n_data=None):
+    def next(self, n_minibatch=None):
```
dustinvtran: Maybe in the same way that the distribution objects sample with a number given by `size`, `data.next()` should use have its argument be called `size`. My guess is that the distinction is that `size` is used generically for any context, and `n_minibatch`/`n_samples` is used specifically in the context of inference and criticism.

edward/models/distributions.py
```diff
@@ -81,7 +81,7 @@ def log_prob(self, xs):
         Parameters
         ----------
         xs : tf.Tensor or np.ndarray
-            n_minibatch x self.shape
+            n_samples x self.shape
```
dustinvtran: In the most generic context (any of use distribution objects), maybe this one should be renamed to `size` similar to `Distribution().sample()`?

edward/models/models.py
```diff
@@ -92,9 +92,9 @@ def _py_log_prob_args(self, zs, *args):
         for key, value in zip(self.keys, values):
             key.set_value(value)
 
-        n_minibatch = zs.shape[0]
-        lp = np.zeros(n_minibatch, dtype=np.float32)
-        for s in range(n_minibatch):
+        n_samples = zs.shape[0]
```
dustinvtran: see above comment in `Distribution().log_prob()`

edward/models/models.py
```diff
@@ -353,8 +353,8 @@ def log_prob(self, xs):
         if len(self.layers) == 1:
             return self.layers[0].log_prob(xs)
 
-        n_minibatch = get_dims(xs[0])[0]
-        log_prob = tf.zeros([n_minibatch], dtype=tf.float32)
+        n_samples = get_dims(xs[0])[0]
```
dustinvtran: see above comment in `Distribution().log_prob()`

examples/bayesian_linear_regression.py
```diff
@@ -38,35 +38,35 @@ class LinearModel:
...
         b = zs[:, 1]
         mus = tf.matmul(x, W) + b
-        # broadcasting to do mus - y (n_data x n_minibatch - n_data x 1)
+        # broadcasting to do mus - y (n_minibatch x n_samples - n_minibatch x 1)
         y = tf.expand_dims(y, 1)
         log_lik = -tf.reduce_sum(tf.pow(mus - y, 2), 0) / self.lik_variance
         return log_lik + log_prior
 
-def build_toy_dataset(n_data=40, noise_std=0.1):
+def build_toy_dataset(n_minibatch=40, noise_std=0.1):
```
dustinvtran: This one refers to the number of data points for building the toy data set. So maybe something like `N`?

examples/bayesian_linear_regression_plot.py
```diff
@@ -39,34 +39,34 @@ class LinearModel:
...
         b = zs[:, 1]
         mus = tf.matmul(x, W) + b
-        # broadcasting to do mus - y (n_data x n_minibatch - n_data x 1)
+        # broadcasting to do mus - y (n_minibatch x n_samples - n_minibatch x 1)
         y = tf.expand_dims(y, 1)
         log_lik = -tf.reduce_sum(tf.pow(mus - y, 2), 0) / self.lik_variance
         return log_lik + log_prior
 
-def build_toy_dataset(n_data=40, noise_std=0.1):
+def build_toy_dataset(n_minibatch=40, noise_std=0.1):
```
dustinvtran: This one refers to the number of data points for building the toy data set. So maybe something like `N`?

examples/bayesian_linear_regression_test.py
```diff
@@ -63,22 +63,22 @@ def predict(self, xs, zs):
         y_pred = tf.reduce_mean(tf.matmul(x_test, W) + b, 1)
         return y_pred
 
-def build_toy_dataset(n_data=40, coeff=np.random.randn(10), noise_std=0.1):
+def build_toy_dataset(n_minibatch=40, coeff=np.random.randn(10), noise_std=0.1):
```
dustinvtran: This one refers to the number of data points for building the toy data set. So maybe something like `N`?

examples/bayesian_nn_analytic_kl.py
```diff
@@ -67,46 +67,46 @@ def mapping(self, x, z):
...
         """Returns a vector [log p(xs | zs[1,:]), ..., log p(xs | zs[S,:])]."""
         x, y = xs['x'], xs['y']
         mus = tf.pack([self.mapping(x, z) for z in tf.unpack(zs)])
-        # broadcasting to do mus - y (n_minibatch x n_data - n_data)
+        # broadcasting to do mus - y (n_samples x n_minibatch - n_minibatch)
         log_lik = -tf.reduce_sum(tf.pow(mus - y, 2), 1) / self.lik_variance
         return log_lik
 
-def build_toy_dataset(n_data=40, noise_std=0.1):
+def build_toy_dataset(n_minibatch=40, noise_std=0.1):
```
dustinvtran: This one refers to the number of data points for building the toy data set. So maybe something like `N`?

examples/hierarchical_logistic_regression.py
```diff
@@ -69,21 +69,21 @@ def log_prob(self, xs, zs):
         log_prior = -self.prior_variance * tf.reduce_sum(zs*zs, 1)
         return log_lik + log_prior
 
-def build_toy_dataset(n_data=40, noise_std=0.1):
+def build_toy_dataset(n_minibatch=40, noise_std=0.1):
```
dustinvtran: This one refers to the number of data points for building the toy data set. So maybe something like `N`?

examples/mixture_gaussian_map.py
```diff
@@ -71,8 +71,8 @@ def log_prob(self, xs, zs):
 
         # Loop over each mini-batch zs[b,:]
```
dustinvtran: `# Loop each sample zs[b, :]`

examples/mixture_gaussian_laplace.py
```diff
@@ -71,8 +71,8 @@ def log_prob(self, xs, zs):
 
         # Loop over each mini-batch zs[b,:]
```
dustinvtran: `# Loop over each sample zs[b, :]`

examples/mixture_gaussian.py
```diff
@@ -67,8 +67,8 @@ def log_prob(self, xs, zs):
 
         # Loop over each mini-batch zs[b,:]
```
dustinvtran: `# Loop over each sample zs[b, :]`

dustinvtran:
Comments added inline. Looks great! I agree with all the naming decisions.

One additional thought: In the distribution object, we use `size` in both `sample()`, `log_prob()`, and `log_prob_idx()`to denote the size from the outer dimension (e.g., number to sample, number of batches, whatever the context it is used in). For the distribution object, maybe we should call it `n` instead to conform to `tf.contrib.distributions`: see, e.g., https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/gamma.py#L332

mariru:
Thanks for the comments @dustinvtran. I included them.

RE distribution object. I think `size` is good but would be happy to change to `n`. What do you prefer?

dustinvtran:
Perfect. Yeah, let's change it to `n`. Also: in `tests/test-models/` we use `n_minibatch` in some places; these should be changed to `n` now as well.

tests/test-stats/test_nbinom_rvs.py
```diff
@@ -11,9 +11,9 @@
 sess = tf.Session()
 
 
-def _test(n, p, size):
-    val_est = nbinom.rvs(n, p, size=size).shape
-    val_true = (size, ) + np.asarray(n).shape
+def _test(n, p, n):
```
dustinvtran: For `test-stats/` and the `edward.stats` library in general, I think we can leave it to be called `size`. That is, `edward.stats` follows the convention of `scipy.stats`. `edward.models` uses `n` because `edward.models` follows the convention of `tf.contrib.distributions`.

dustinvtran:
Hi @mariru. I made slight changes to deal with merge conflicts. This looks good to me. Is there anything else to do before I merge this?

dustinvtran:
To outline the convention,

+ Total number of data points: `N`
+ Total number of features/covariates: `D`
+ Total number of clusters/any other auxiliary concept to number: `K`

+ Batch size of data during training (if inference uses batches): `n_minibatch`
+ Batch size of latent variable samples during training (if inference uses latent variable samples): `n_samples`

+ Number of random variables: `n_vars`
+ Number of parameters: `n_params`
+ Number of samples to sample in `edward.models`: `n` (following `tf.contrib.distributions)
+ Number of samples to sample in `edward.stats`: `size` (following `scipy.stats`)

mariru:
Looks good @dustinvtran

dustinvtran:
Great! merging now.

Issue #137, opened 2016-06-28
Distributed training example + tests


jaanli:
Distributed training is a crucial feature. We need to make sure that at least all the examples run in a distributed way in the tests (multiple CPUs / multiple GPUs).

dustinvtran:
I agree. I haven't checked it but my guess is that you can just wrap everything inside a device context, whether it be CPU or GPU. (It is all just defining a computational graph in TensorFlow and running an optimizer.) Batch training will be a different story as it depends on the inference algorithm.

PR #170, opened 2016-07-12
Feature/data reader

### LINKED ISSUE #92: random data subsampling (in addition to slices/windows)
this PR https://github.com/blei-lab/edward/pull/90/ implements a sliding window style of subsampling.

do we also want to implement random subsampling? if so, do we need any architectural changes to `data.py`? (i don't think so.)

dustinvtran:
#### Summary:

+ issues: fixes #92 (one can change the arguments to `tf.train.batch`); related to #137

This pull request comes at the advice of the Bayesflow folks to support data readers.

There are now three ways to read data in Edward, following the
[three ways to read data in TensorFlow](https://www.tensorflow.org/versions/r0.9/how_tos/reading_data/index.html).

1. __Preloaded data.__ A constant or variable in the TensorFlow graph holds all the data.

   For inference, pass in the data as a dictionary of NumPy arrays. Internally, we will store them in TensorFlow variables to prevent copying data more than once in memory. Batch training is available internally via `tf.train.slice_input_producer` and `tf.train.batch`. (As an example, see `examples/mixture_gaussian.py`.)
2. __Feeding.__ Manual code provides the data when running each step of inference.

   For inference, pass in the data as a dictionary of TensorFlow placeholders. The user must manually feed the placeholders at each step of inference. (As an example, see `examples/mixture_density_network.py` or `examples/convolutional_vae.py`.)
3. __Reading from files.__ An input pipeline reads the data from files at the beginning of a TensorFlow graph.

   For inference, pass in the data as a dictionary of TensorFlow tensors, where the tensors are the output of data readers. (No current example is available.)

Internally, there is no longer a need for a `DataGenerator` object on NumPy arrays because we do the batch training directly on TensorFlow tensors. This simplifies a lot of the code. 

One additional consequence of this pull request is that inference always uses a coordinator and queue runner for distributed training.

#### Intended Effect:

See above.

#### How to Verify:

Run examples referred to above.

#### Side Effects:

None.

#### Documentation:

I added and clarified tthe documentation.

#### Reviewer Suggestions:

@altosaar because we discussed data readers at length. @heywhoah because this started based on the discussion of how to load the data into the computational graph if it fit in memory.


edward/inferences.py
```diff
@@ -180,16 +178,24 @@ def initialize(self, n_iter=1000, n_data=None, n_print=100,
...
 
         loss = self.build_loss()
         if optimizer is None:
             var_list = tf.get_collection(tf.GraphKeys.TRAINABLE_VARIABLES,
                                          scope=scope)
-            # Use ADAM with a decaying scale factor
+            # Use ADAM with a decaying scale factor.
             global_step = tf.Variable(0, trainable=False)
             starter_learning_rate = 0.1
             learning_rate = tf.train.exponential_decay(starter_learning_rate,
```
altosaar: do we need a decaying learning rate?

edward/inferences.py
```diff
@@ -208,6 +214,10 @@ def initialize(self, n_iter=1000, n_data=None, n_print=100,
         init = tf.initialize_all_variables()
         init.run()
 
+        # Start input enqueue threads.
```
altosaar: is it possible to make this optional or have some stat for whether there are any queue runners? otherwise if the user escapes (ctrl+c) we'll have to catch that and stop the threads, e.g.
```
# Start input enqueue threads.
coord = tf.train.Coordinator()
threads = tf.train.start_queue_runners(sess=sess, coord=coord)
# training ops
...
# if training interrupted, or escape sequence
coord.request_stop()
coord.join(threads)
sess.close()
```

edward/inferences.py
dustinvtran: I don't think so. We've seen it help in practice for convergence on toy problems in the `examples/` directory. We can change it in another pull request during heavier investigation.

edward/inferences.py
```diff
@@ -238,14 +246,14 @@ def print_progress(self, t, loss):
                 print(self.variational)
 
     def finalize(self):
```
altosaar: again, i'd make this optional based on whether feed_dicts are being used. starting and stopping threads all the time is not ideal for fast debugging

edward/inferences.py
dustinvtran: The queue runners are necessary for batch training if the data fits in memory (we use `tf.train.slice_input_producer` and `tf.train.batch`, which in turn uses queues). And in general for distributed training when using data readers. 

So it sounds like we should build a method for catching ctrl-c/session interrupts?

edward/inferences.py
```diff
@@ -180,16 +186,24 @@ def initialize(self, n_iter=1000, n_data=None, n_print=100,
...
 
-        # Set shape of data placeholders according to batch size.
-        if not isinstance(self.model, StanModel):
-            for value in six.itervalues(self.data):
-                value.set_shape([self.n_data] + get_dims(value)[1:])
+        if n_data is not None and not isinstance(self.model, StanModel):
+            # Re-assign data to batch tensors, with size given by `n_data`.
+            values = list(six.itervalues(self.data))
+            slices = tf.train.slice_input_producer(values)
+            batches = tf.train.batch(slices, self.n_data)
```
dustinvtran: @altosaar mentioned via Slack that we should default to a larger number of threads than 1.

dustinvtran: @altosaar mentioned via Slack that we should default to a larger number of threads than 1.

edward/inferences.py
dustinvtran: If the user is in the setting of preloaded data or reading from files, it could still be useful to continue using the same threads after interrupting computation.

I guess if there are _no_ queues we could have it not use a coordinator and queue runner. Tracking this would require storing an additional class attribute that tracks whether any of the data values are not placeholders. I think this could add a complication that may lead to harder debugging as one must keep this in mind. It's easier to think of the default of _always_ using a queue runner.

edward/inferences.py
dustinvtran: If the user is in the setting of preloaded data or reading from files, it could still be useful to continue using the same threads after interrupting computation. So let's not do my proposal above.

dustinvtran:
from @altosaar on Slack:
> just took a look - where can i check whether it passes the unit test for data?
> (i'm guessing this is it https://github.com/blei-lab/edward/blob/master/tests/test_data_next.py) (edited)
> maybe it's worth having a speed check? have you run this e.g. on the VAE example to see if there's a loss in speed?
> i think speed / iterations per second is something that would be useful to track/print in general along with the ELBO, etc
> this way users can see how much of a speedup they get on GPUs
> and if queuerunners are slowing things down

it doesn’t pass the unit test for data because the unit test (and data object) no longer exists :)

a speed check is worth doing. it shouldn’t make a difference for starting a queue runner if there are no queues

re:tracking speed / iterations per sec. yeah that sounds very useful. can you add that as a github issue?

altosaar:
can you please add a unit test equivalent to https://github.com/blei-lab/edward/blob/master/tests/test_data_next.py to make sure batch / queuerunners are working as expected?

dustinvtran:
@altosaar: added

altosaar:
@dustinvtran awesome! looks great. 

can we also add a test of reading a simple .tfrecords file? 

this is the crucial test for ensuring that the queuerunners are being started and stopped properly.

dustinvtran:
We know that the queue runners are already being started and stopped properly because `test_preloaded_batch()` is using it.

I can look into the tfrecords though. Do you know a good reference implementation?

altosaar:
I'd try the example from indico but with a small synthetic dataset for the model (`writer = tf.python_io.TFRecordWriter("mnist.tfrecords")`): https://indico.io/blog/tensorflow-data-inputs-part1-placeholders-protobufs-queues/

dustinvtran:
@altosaar: Done! Feel free to merge if you think it's ready now :)

tests/test_inference_data.py
```diff
@@ -10,7 +10,6 @@
 from edward.models import Variational, Normal
 from edward.stats import norm
 
-
 ed.set_seed(1512351)
```
altosaar: shouldn't we be randomizing the seed?

PR #175, opened 2016-07-16
get shape in util.py function hessian() by evaluating tf tensor

mariru:
necessary fix to get 
`examples/mixture_gaussian_laplace.py` to work.
Not sure if this is the most efficient way to get the shape but it works.

Just changed one line.
Before `grads.get_shape()[0]` was returning `?` in the example. Now the tensor is evaluated to get the shape.

### DIFF
--- a/edward/util.py
+++ b/edward/util.py
@@ -128,7 +128,9 @@ def hessian(y, xs):
     grads = tf.concat(0, grads)
     # Loop over each element in the vector.
     mat = []
-    d = grads.eval().shape[0]
+    d = grads.get_shape()[0]
+    if type(d) != int:
+        d = grads.eval().shape[0]
     for j in range(d):
         # Calculate grad_{xs} ( [ grad_{xs} y ]_j ).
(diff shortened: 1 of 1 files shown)


dustinvtran:
Thanks! The `mixture_gaussian_laplace.py` works on my end though. The `grads` and `d` are `Tensor("concat_2:0", shape=(10,), dtype=float32)` and `10` respectively. 

In general `grads` is the output of evaluating the gradient of `y` with respect to each `x` in `xs`, then flattened and concatenated. I guess generally the shape of `grads` can be unknown if any shape of an `x` in `xs` is unknown? So it sounds useful. To check types, can you use `isinstance(d, int)`? (This is the PEP 8 way.)

mariru:
Weird! For me `grads` is `Tensor("concat_2:0", shape=(?,), dtype=float32)`

Changed the type chacking as suggested.

dustinvtran:
And for PEP 8, a space between between the comma and `d` :) (feel free to `git commit --amend` and `git push -f` to avoid such a trivial git commit)

PR #182, opened 2016-07-19
inference network tutorial

diengadji:
This PR does the following:
1 - tutorial on VAE with a particular focus on the encoder part

Side Effects:
N\A

Reviewer Suggestions:
@dustinvtran

### DIFF
--- a/docs/tex/tut_inference_network.tex
+++ b/docs/tex/tut_inference_network.tex
@@ -1,0 +1,92 @@
+\documentclass{article}
+\usepackage{amsmath}
+\usepackage{hyperref}
+\usepackage{listings}
+\usepackage{tikz}
+\usetikzlibrary{bayesnet}
+
+% Define the subtitle of the page
+\title{Variational Autoencoders}
+
+
+\begin{document}
+% Begin the content of the page
+\subsection*{Variational Autoencoders}
+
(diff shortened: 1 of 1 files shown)


diengadji:
There is a small typo about the decoder after the equation for the ELBO ...it should be p(X | z; \theta) instead of p(z | X; theta)

dustinvtran:
Hi @diengadji. The inference network tutorial is more about how do we use neural networks (or generally, any function) to amortize inference? It goes into the inference section and does not reference a specific model example, but goes into how to build parameterized variational distributions in Edward and why inference networks are useful.

This looks like a good variational auto-encoder tutorial for the end-to-end section, which sweeps through the whole process. Can you frame it in the same structure as the other tutorials in this section? That is, brief intro, then Model, Inference, Criticism, done.

dustinvtran:
I wrote a inference networks tutorial here: http://edwardlib.org/tut_inference_networks

The tutorial in this pull request looks like a great end-to-end example which uses the probabilistic decoders tutorial, the inference networks tutorial, and the (currently unwritten) prior predictive checks tutorial. Would be great if it's structured in that sort of format.

diengadji:
Great inference network tutorial! I will reorganize this one to be an end-to-end VAE tutorial. Thanks.

PR #119, opened 2016-06-10
[WIP] Fix for #74

kudkudak:
Otherwise this happens :) :

```{python}
import edward
from edward.models import Variational
from edward.models import Normal

print Variational().layers
Variational().add(Normal(2))
print Variational().layers
```

prints

```
[]
[<edward.models.variationals.Normal instance at 0x117cc0680>]
```

Issue #74, opened 2016-05-09
Running example model twice


twiecki:
If I run the beta-bernoulli model more than once, I get:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-b5d0bc03f642> in <module>()
     29 inference = ed.MFVI(model, variational, data)
     30 
---> 31 inference.run(n_iter=10000)

/home/wiecki/working/projects/edward/edward/inferences.py in run(self, *args, **kwargs)
     64         sess = self.initialize(*args, **kwargs)
     65         for t in range(self.n_iter):
---> 66             loss = self.update(sess)
     67             self.print_progress(t, loss, sess)
     68 

/home/wiecki/working/projects/edward/edward/inferences.py in update(self, sess)
    162             samples = self.variational.sample_noise(self.samples.get_shape())
    163 
--> 164         _, loss = sess.run([self.train, self.losses], {self.samples: samples})
    165 
    166         return loss

/home/wiecki/miniconda3/lib/python3.5/site-packages/tensorflow/python/client/session.py in run(self, fetches, feed_dict, options, run_metadata)
    338     try:
    339       result = self._run(None, fetches, feed_dict, options_ptr,
--> 340                          run_metadata_ptr)
    341       if run_metadata:
    342         proto_data = tf_session.TF_GetBuffer(run_metadata_ptr)

/home/wiecki/miniconda3/lib/python3.5/site-packages/tensorflow/python/client/session.py in _run(self, handle, fetches, feed_dict, options, run_metadata)
    551                 'Cannot feed value of shape %r for Tensor %r, '
    552                 'which has shape %r'
--> 553                 % (np_val.shape, subfeed_t.name, str(subfeed_t.get_shape())))
    554           if not self.graph.is_feedable(subfeed_t):
    555             raise ValueError('Tensor %s may not be fed.' % subfeed_t)

ValueError: Cannot feed value of shape (1, 5) for Tensor 'samples_4:0', which has shape '(1, 1)'
```

The counter seems to increase each time I execute.

akucukelbir:
thanks @twiecki.

does https://github.com/blei-lab/edward/issues/70 solve it?

kudkudak:
[...]

PR #184, opened 2016-07-24
Forcing user to correctly seed code

kudkudak:
**Summary**

Due to holding single TF session code is not properly seeded. If you run example in a loop twice (with seeding) it will produce different results. What's maybe worse, if user places ``ed.set_seed()`` too late in a single run code (after creation of at least one random variable) it will also be stochastic.

There is no simple way of fixing that, simplest way I could think of was creating some auxiliary class for storing what initializers were used for Variables and recreating all Variables in the graph when running ``ed.set_seed()``.

Instead I propose to force user to place ``ed.set_seed()`` before creating any variables. Now code produces a RuntimeError if user does differently.

(This is continuation of https://github.com/blei-lab/edward/pull/119 and #74)

Reference on seeding in TF: https://www.tensorflow.org/versions/r0.8/api_docs/python/constant_op.html#set_random_seed

**How to verify**

This snippet now throws error instead of producing incosistent results despite seeding
```
class BetaBernoulli:
    """p(x, z) = Bernoulli(x | z) * Beta(z | 1, 1)"""
    def log_prob(self, xs, zs):
        log_prior = beta.logpdf(zs, a=1.0, b=1.0)
        log_lik = tf.pack([tf.reduce_sum(bernoulli.logpmf(xs['x'], z))
                           for z in tf.unpack(zs)])
        return log_lik + log_prior


model = BetaBernoulli()
variational = Variational()
variational.add(Beta())
data = {'x': np.array([0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}

ed.set_seed(42)
inference = ed.MFVI(model, variational, data)
inference.run(n_iter=10)
```

**Reviewer Suggestions**
@dustinvtran

### DIFF
--- a/edward/util.py
+++ b/edward/util.py
@@ -440,6 +440,10 @@ def set_seed(x):
     x : int, float
         seed
     """
+    if len(tf.get_default_graph()._nodes_by_id.keys()) > 0:
+        raise RuntimeError("Seeding is not supported after initializing part of the graph. "
+                           "Please move set_seed to the beginning of your code.")
+
     np.random.seed(x)
     tf.set_random_seed(x)
 


dustinvtran:
Nice! This is elegant. I guess the difficulty is correctly framing the problem, and the solution sort of pops out. 

Let me summarize to see if I understand correctly. So if any operations with randomness are constructed in the graph—_before_ setting the graph-level seed (`tf.set_random_seed()`)—then we won't be able to reproduce the randomness in those operations. This occurs in Edward for example when we create the variational model before calling `ed.set_seed()`: the variational model creates TensorFlow variables, which uses random ops (it initializes with `tf.random_normal()`).

It sounds like `ed.set_seed()` must go not only before any TensorFlow variables are created, but before any TensorFlow random operations are created. This is ideal. However, traversing the current graph and checking if each op is random sounds expensive. So we can just be conservative and check that the TensorFlow graph is empty before calling `ed.set_seed()`.
```python
if len(tf.get_default_graph()._nodes_by_id.keys()) > 0:
    raise RuntimeError("Seeding is not supported after building part of the graph. "
                       "Please move set_seed to the beginning of your code.")
```

kudkudak:
> Let me summarize to see if I understand correctly. So if any operations with randomness are constructed in the graph—before setting the graph-level seed (tf.set_random_seed())—then we won't be able to reproduce the randomness in those operations. This occurs in Edward for example when we create the variational model before calling ed.set_seed(): the variational model creates TensorFlow variables, which uses random ops (it initializes with tf.random_normal()).

Exactly

> It sounds like ed.set_seed() must go not necessarily before any TensorFlow variables are created, but before any TensorFlow random operations are created. 

Right, thanks for catching that. This unfortunately makes even more tests fail :P I will look into this this week and keep PR open as it shouldn't take too long.

dustinvtran:
One thought to avoid failing tests is to have `ed.set_seed()` raise a warning instead of raise an error.

kudkudak:
@dustinvtran Well it is by design raising error - otherwise it is very easy to skip warning, and I thought that seeding error is very dangerous. Obviously, I leave it to you to decide, should we change it to warning?

dustinvtran:
I can see it either way. Up to you!

kudkudak:
@dustinvtran So I think it is ready for merge :) Added tf.reset_default_graph() in tests, because nosetests run all the tests within one process. This is slightly tricky, so maybe worth documenting somewhere?

dustinvtran:
@kudkudak apologies for the delay. 

We recently switched to using [py.test](http://doc.pytest.org/en/latest/) (more details on the [contributing page](http://edwardlib.org/contributing)). I was able to run py.test successfully without the `tf.reset_default_graph()` calls, e.g., with
```python
py.test tests/test-models/*
```
So that all sounds good! Can you verify this works on your end too? Then we can go ahead with rebasing the commits all as one commit and merge this.

edward/util.py
```diff
@@ -434,12 +434,16 @@ def rbf(x, y=0.0, sigma=1.0, l=1.0):
...
 def set_seed(x):
     """Set seed for both NumPy and TensorFlow.
-
     Parameters
     ----------
     x : int, float
         seed
     """
+    if len(tf.get_default_graph()._nodes_by_id.keys()) > 0:
+        from nose.tools import set_trace; set_trace()
```
dustinvtran: can you remove this import line, and also undelete the line that was deleted above?

kudkudak: ! Sorry :)

dustinvtran: no problem! all is fair in the land of experimentation

kudkudak:
@dustinvtran Cool, thanks for the catch. Rebased and addressed your comment.

dustinvtran:
Thanks again!

PR #187, opened 2016-07-27
Feature/util asserts

### LINKED ISSUE #60: write tests for functions in edward/util.py
there are many small functions in `edward/util.py` that we depend on.

let's write unit tests for them.

### LINKED ISSUE #147: linter
we may want to set up a linter, both for code and docstring (numpy style) checking.

### LINKED ISSUE #186: convert unit tests to tensorflow testing format
https://www.tensorflow.org/versions/master/api_docs/python/test.html

akucukelbir:
#### Summary:

+ issues: fix #60; fix #147; fix #186

We introduce input checking to functions under `util.py`. We update tests to reflect this. 

We correct some inconsistencies in functions in `util.py` and correct minor typos in docstrings.

We do not check for shapes (yet).

#### Intended Effect:

Make building blocks more robust.

#### How to Verify:

Look at the changes. Run new tests.

#### Side Effects:

N/A

#### Documentation:

N/A

#### Reviewer Suggestions:

@dustinvtran


dustinvtran:
Let me summarize our discussion yesterday (it personally helps me organize my thoughts): there's a distinction between the many things that need to be checked. One is the distinction between checks at construction time (i.e., `f(x)` raises an error when calling it, such as when inputs have wrong type) and checks at run time (e.g., `sess.run(f(x))` raises an error, such as when tensors suddenly become nan's/inf's). The assert ops are for checking at run time. Further, we make the design decision that we only check inputs, i.e., not the body of a function or its output.

The second thing we want to do is TensorFlow-ize our unit testing. This adds more functionality to our unit tests and lets us avoid reinventing the wheel.

This all sounds good. Now the last point is to decide on the best way to incorporate these assert ops. This pull requests takes the convention of first declaring `assert_ops`, then wrapping the function body in an `assert_ops` context. We'd like to appeal to other authorities. But our search for robust Python functions implemented in TensorFlow made us realize that there is probably no convention.

dustinvtran:
Okay, so `tf.contrib.distributions` has a convention for assert ops in class objects. They define a private function, `_check_x(self, x)`, which runs assert ops on `x`. They then call `self._check_x` to apply them to any relevant inputs in all class methods. [Here's an example](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/beta.py#L305). This makes sense because if the assertions are only checking inputs, then it doesn't make sense to wrap the function body within the assertion; it should be applied before any inputs are used.

For functions, maybe we can do something like:
```python
def f(x):
    dependencies = [
        check_ops.assert_positive(x),
        check_ops.assert_less(x, constant_op.constant(
1, self.dtype))] if self.validate_args else []
    x = control_flow_ops.with_dependencies(dependencies, x)
    # function body here
```
Thoughts?

akucukelbir:
@dustinvtran i made the switch for all the functions, except `hessian`. i couldn't get the mystery magic function to work for a list of tensors of different shapes. (the old formulation still works though.)

edward/util.py
```diff
@@ -179,16 +206,31 @@ def kl_multivariate_normal(loc_one, scale_one, loc_two=0, scale_two=1):
...
+    if loc_two == 0.0 and scale_two == 1.0:
         return 0.5 * tf.reduce_sum(
             tf.square(scale_one) + tf.square(loc_one) - \
             1.0 - 2.0 * tf.log(scale_one))
     else:
         return 0.5 * tf.reduce_sum(
             tf.square(scale_one/scale_two) + \
             tf.square((loc_two - loc_one)/scale_two) - \
-            1.0 + 2.0 * tf.log(scale_two) - 2.0 * tf.log(scale_one), 1)
+            1.0 + 2.0 * tf.log(scale_two) - 2.0 * tf.log(scale_one))
```
dustinvtran: Seems like this one modifies the behavior?

edward/util.py
```diff
@@ -205,9 +247,17 @@ def log_mean_exp(x):
...
+    ------
+    InvalidArgumentError
+        If the input has Inf or NaN values.
     """
+    dependencies = [tf.verify_tensor_all_finite(x, msg='')]
+    x = control_flow_ops.with_dependencies(dependencies, x)
+
     x_max = tf.reduce_max(x)
-    return tf.add(x_max, tf.log(tf.reduce_mean(tf.exp(tf.sub(x, x_max)))))
+    return tf.add(x_max, tf.log(tf.reduce_mean(tf.exp(tf.sub(x, x_max)))))
```
dustinvtran: trailling whitespace!

edward/util.py
```diff
@@ -243,8 +299,16 @@ def logit(x):
...
     tf.Tensor
         size corresponding to size of input
+
+    Raises
+    ------
+    InvalidArgumentError
+        If the input is not between :math:`(0,1)` elementwise.
     """
-    x = tf.clip_by_value(x, 1e-8, 1.0 - 1e-8)
+    dependencies = [tf.assert_positive(x),
```
dustinvtran: This is an interesting decision, i.e., to not do any clipping. I think I agree with it.

edward/util.py
```diff
@@ -268,16 +332,31 @@ def multivariate_rbf(x, y=0.0, sigma=1.0, l=1.0):
...
+                    tf.assert_positive(l)]
+    x = control_flow_ops.with_dependencies(dependencies, x)
+    y = control_flow_ops.with_dependencies(dependencies, y)
+    sigma = control_flow_ops.with_dependencies(dependencies, sigma)
+    l = control_flow_ops.with_dependencies(dependencies, l)
+
     return tf.pow(sigma, 2.0) * \
-           tf.exp(-1.0/(2.0*tf.pow(l, 2.0)) * \
-                  tf.reduce_sum(tf.pow(x - y , 2.0)))
+            tf.exp(-1.0/(2.0*tf.pow(l, 2.0)) * \
```
dustinvtran: whitespace!

edward/util.py
```diff
@@ -357,17 +471,25 @@ def to_simplex(x):
     Parameters
     ----------
     x : tf.tensor or np.array
-        scalar, vector, or matrix
+        vector, or matrix
```
dustinvtran: this should work for scalar as well (i.e., all tf.tensors). we can do a sweep in another pull request changing how we write documentation of tensors; see #191

tests/test-util/test_get_dims.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def test_get_dims_2d_np(self):
+        with self.test_session():
+            x = np.zeros([2,2])
+            self.assertEqual(get_dims(x), [2,2])  
+
+if __name__ == '__main__':
+    tf.test.main()
+
+
```
dustinvtran: trailing lines!

dustinvtran:
Hi @akucukelbir. All looks good and ready to merge. What should we do about the failing Travis build and possibly moving to py.test?

akucukelbir:
- [ ] manually run examples before merging, just as a sanity check.

dustinvtran:
Everything passed. We will deal with PEP 8 in a separate pull request. We will deal with testing `kl_multivariate_normal` in a separate pull request. Merging now.

PR #197, opened 2016-08-02
fix pymc3model bug

dustinvtran:
Recent installations of PyMC3 will not work with our `PyMC3Model` wrapper. It errors with `inputvars` not found in the PyMC3 module.

This pull request (https://github.com/pymc-devs/pymc3/pull/1257) removed `core.py`. It used to import to `__init__.py` the method `inputvars`  via `theanof.py`, and the method `DictToArrayBijection` and `ArrayOrdering` via `blocking.py`.

In this pull request, I manually added all the required methods via copy-pastes. `PyMC3Model` now works. But I don't suggest we do this. This pull request will be useful to discuss what our next steps should be. Perhaps we should contact the PyMC3 devs on their API plan.

### DIFF
--- a/edward/models/models.py
+++ b/edward/models/models.py
@@ -16,7 +16,6 @@ except ImportError:
     pass
 
 try:
-    #import pymc3 as pm
     from theano import theano, scalar, tensor as tt
     from theano.gof.graph import inputs
     def makeiter(a):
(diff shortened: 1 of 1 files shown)


dustinvtran:
@akucukelbir and I discussed. We are merging this hack to get the model wrapper working for now. We contacted PyMC3 devs and will wait to hear back from them.

PR #202, opened 2016-08-04
Fixes integer truncation for BetaBern example.

tshauck:
It appears that because of integer dataset in the Beta Bernoulli example
that `ed.ppc` will return the mean of the dataset, but truncated to an
integer.  As such, I'm getting 1s in the event each sample was a 1 and
0s otherwise.

I'd be happy to update the docs (such as the delving in) to reflect this change,
but I wasn't sure if making the change to the docs would be worth the added
mental challenge of thinking about integer math.

### DIFF
--- a/examples/beta_bernoulli_ppc.py
+++ b/examples/beta_bernoulli_ppc.py
@@ -50,5 +50,5 @@ data = {'x': np.array([0, 1, 0, 0, 0, 0, 0, 0, 0, 1])}
 inference = ed.MFVI(model, variational, data)
 inference.run(n_iter=200)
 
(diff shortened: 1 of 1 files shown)


tshauck:
Failure looks to be PEP8 related, and not this code change.

dustinvtran:
Good catch! Also you're right in the decision not to change the docs. This seems specific to the fact that the data's NumPy array has type int and not float. We've tried to be careful about this in the codebase by casting everything to float32 internally, although inevitably some examples do the type casting manually. For now, I'm happy to merge this pull request. In a separate pull request, we can make consistent the type casting in all Edward functions (#203).

PR #205, opened 2016-08-04
fix mixture gaussian's log joint density

### LINKED ISSUE #185: Gaussian mixture example
I've been trying to experiment a bit with the GMM example, drawing a scatter plot etc., but I can't get the model to actually work (code: https://gist.github.com/DarkSensei/b7d31ff7d3d38633df11ab6316865b9a) . I did some poking around, and the only way I could get the model to learn, is by setting the variances to be unit, otherwise they just explode into large numbers. I'd appreciate if someone could lend a hand with this.

dustinvtran:
This fixes #185. Specifically, we fix the mixture of Gaussian's density by expanding upon our log sum exp utility function. I also prefer @fmpr's toy dataset over a pre-simulated text file—the former is more transparent and has visualizations—so I replaced the dataset. Reference example to look at is the `examples/mixture_gaussian.py` script.

Comments welcome. @DarkSensei and @fmpr: Let me know if this one is more helpful. 

Eventually, I'd like to include some visualizations into the "Unsupervised learning" tutorial. For example, maybe we can generate new datasets and see how much it looks like the real dataset. Or we can visualize the predicted cluster memberships for each data point. Either it goes into this pull request or someone can work it on after this one is merged?


fmpr:
Great! Much more numerically stable than my naive implementation. Thanks!

We can compute the most likely cluster assignments by including the following predict function, which  uses of the posterior mean of the latent variables:

    def predict(self, xs, zs):
        """Return a cluster assignment for each data point, averaging over
        each set of latent variables z in zs."""
        x = xs['x']
        pi, mus, sigmas = zs
        pi = tf.reduce_mean(pi, 0)
        mus = tf.reduce_mean(mus, 0)
        sigmas = tf.reduce_mean(sigmas, 0)

        matrix = []
        for k in range(self.K):
            matrix += [multivariate_normal.logpdf(x,
                               mus[(k*self.D):((k+1)*self.D)],
                               sigmas[(k*self.D):((k+1)*self.D)])]

        return tf.pack(matrix)
and visualizing the predicted cluster memberships using:
```
clusters = np.argmax(ed.evaluate('log_likelihood', model, variational, data), axis=0)
plt.scatter(data['x'][:, 0], data['x'][:, 1], c=clusters, cmap=cm.bwr)
plt.show()
```

I guess then it would not make sense to plot the true cluster assignments in the `build_toy_dataset()` function at the beginning... Maybe just plot the raw data using:
```
plt.scatter(data['x'][:, 0], data['x'][:, 1])
plt.show()
```

dustinvtran:
Perfect! So if I understand correctly, with `predict()` you output a K x N matrix of log-likelihoods, per data-point and per-cluster assignment. Then you choose the predicted assignment based on argmax along the rows, i.e., the cluster assignment which produces the highest log-likelihood for each data point.

I'd like you to get contributor acknowledgement. I'll merge this now. How about you include the above method and lines of code into `mixture_gaussian.py` in a separate pull request? I can also help with getting the outputted visualizations into the unsupervised learning tutorial (`docs/tex/tut_unsupervised.tex`).

fmpr:
Exactly! That is precisely what the `predict()` function is doing. Using the artificial dataset of the example it recovers the correct cluster assignments.

Sure, I can make a separate pull request with this code. But do I have permissions for that?

dustinvtran:
Yep, you can fork the repo, make changes on your fork, and then submit a PR to merge your branch to the one here.

PR #208, opened 2016-08-05
add predict() function for mixture_gaussian.py example

fmpr:
This improves the mixture_gaussian.py example by adding a predict() function, in order to estimate the most likely cluster assignments based on the (log)likelihood of the data point under the different gaussians. A visualization of the estimated cluster assignments for the artificial dataset of the example shows that it recovers the correct assignments.

### DIFF
--- a/examples/mixture_gaussian.py
+++ b/examples/mixture_gaussian.py
@@ -98,24 +98,38 @@ class MixtureGaussian:
 
         return log_prior + tf.pack(log_lik)
 
+    def predict(self, xs, zs):
+        """Return matrix with log-likelihoods for each data point under each cluster, 
+        averaging over each set of latent variables z in zs."""
+        x = xs['x']
+        pi, mus, sigmas = zs
+        pi = tf.reduce_mean(pi, 0)
+        mus = tf.reduce_mean(mus, 0)
+        sigmas = tf.reduce_mean(sigmas, 0)
+
+        matrix = []
+        for k in range(self.K):
+            matrix += [multivariate_normal.logpdf(x,
+                               mus[(k*self.D):((k+1)*self.D)],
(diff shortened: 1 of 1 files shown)


examples/mixture_gaussian.py
```diff
@@ -98,24 +98,38 @@ def log_prob(self, xs, zs):
...
+        mus = tf.reduce_mean(mus, 0)
+        sigmas = tf.reduce_mean(sigmas, 0)
+
+        matrix = []
+        for k in range(self.K):
+            matrix += [multivariate_normal.logpdf(x,
+                               mus[(k*self.D):((k+1)*self.D)],
+                               sigmas[(k*self.D):((k+1)*self.D)])]
+
+        return tf.pack(matrix)
```
dustinvtran: For PEP 8 reasons, can you add an additional blank line here after this return statement? That is, [surround top-level function and class definitions with two blank lines.](https://www.python.org/dev/peps/pep-0008/#blank-lines)

examples/mixture_gaussian.py
```diff
@@ -98,24 +98,38 @@ def log_prob(self, xs, zs):
...
         x[n, :] = np.random.multivariate_normal(mus[k], np.diag(stds[k]))
-        plt.scatter(x[n, 0], x[n, 1], color=colors[k])
-
-    plt.show()
     return {'x': x}
 
 
 ed.set_seed(42)
 data = build_toy_dataset(500)
+plt.scatter(data['x'][:, 0], data['x'][:, 1])
```
dustinvtran: Can you add a title to this plot? (and also the plot after inference)

dustinvtran:
Thanks! I'll merge this now, and then will incorporate the additional code snippets into the unsupervised learning tutorial.

fmpr:
Great!

PR #215, opened 2016-08-12
init pep8

akucukelbir:
#### Summary:

+ converts `__init__.py` to pass PEP8

#### Intended Effect:

See above.

#### How to Verify:

Look at code. Run tests.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions:

@dustinvtran

### DIFF
--- a/edward/__init__.py
+++ b/edward/__init__.py
@@ -11,5 +11,8 @@ from edward import util
 # Direct imports for convenience
 from edward.models import PyMC3Model, PythonModel, StanModel
 from edward.criticisms import evaluate, ppc
-from edward.inferences import Inference, MonteCarlo, VariationalInference, MFVI, KLpq, MAP, Laplace
-from edward.util import cumprod, dot, get_dims, get_session, hessian, kl_multivariate_normal, log_sum_exp, logit, multivariate_rbf, rbf, set_seed, softplus, stop_gradient, to_simplex
+from edward.inferences import Inference, MonteCarlo, VariationalInference, \
+    MFVI, KLpq, MAP, Laplace
+from edward.util import cumprod, dot, get_dims, get_session, hessian, \
+    kl_multivariate_normal, log_sum_exp, logit, multivariate_rbf, rbf, \
+    set_seed, softplus, stop_gradient, to_simplex


dustinvtran:
It seems like the TensorFlow convention is to always do imports in a single line, and disable PEP8's error for that line. [E.g., `tf.contrib.distributions.Normal`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/normal.py#L23) Is there a way we can do something similar?

dustinvtran:
Following Slack discussion, we will force all code to be <=80 lines (no exceptions).

> alp [12:47 PM]  
> how do you want to proceed
> 
> [12:47]  
> i'm not a huge fan of
> 
> [12:48]  
> `# pylint: disable=line-too-long`
> 
> [12:48]  
> for instance i hardwrap my code
> 
> [12:48]  
> with a shortcut
> 
> [12:48]  
> if i'm not careful, i'll end up hardwrapping this line because it's > 80
> 
> dustin [12:49 PM]  
> yeah that makes sense
> 
> [12:49]  
> if this pylint thing only occurs in this few spots, then maybe we shouldn’t do it.
> 
> [12:50]  
> in general, are there other circumstances where hardwrapping the line would lead to unreadable code?
> 
> alp [12:51 PM]  
> i'm not sure i followed your last two sentences
> 
> dustin [12:52 PM]  
> if there are only these many spots where we want to prevent hardwrapping, then maybe we shouldn’t do it.
> 
> [12:52]  
> in general, are there other circumstances where hardwrapping the line would lead to unreadable code?
> 
> alp [12:53 PM]  
> prevent hardwrapping is an editor thing
> 
> [12:53]  
> you mean doing the `# pylint: disable=line-too-long` hack to avoid pylint from complaining?
> 
> [12:53]  
> oaky
> 
> dustin [12:53 PM]  
> yes
> 
> alp [12:53 PM]  
> let me take a step back and ask a simpler question
> 
> [12:53]  
> do we want all of our code to have < 80
> 
> [12:53]  
> my answer is: yes, i think it should
> 
> dustin [12:54 PM]  
> the answer to that question is the answer to the question above :)
> 
> alp [12:54 PM]  
> great :)
> 
> [12:54]  
> sorry, my brain is running on fumes
> 
> dustin [12:55 PM]  
> okay i checked tensorflow https://github.com/tensorflow/tensorflow/search?utf8=✓&q=disable%3Dline-too-long&type=Code
> 
> [12:55]  
> in all these cases that i looked at, breaking up the line doesn’t impend readability very much
> 
> [12:56]  
> so let’s go with that. and force <80 everwhere
> 
> alp [12:56 PM]  
> ok

PR #217, opened 2016-08-12
Feature/pep8 criticisms

akucukelbir:
#### Summary:

+ converts `criticisms.py` to pass PEP8

#### Intended Effect:

See above.

#### How to Verify:

Look at code. Run tests.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions:

@dustinvtran

### DIFF
--- a/edward/criticisms.py
+++ b/edward/criticisms.py
@@ -10,428 +10,428 @@ from edward.util import logit, get_dims, get_session
 
 
 def evaluate(metrics, model, variational, data, y_true=None, n_samples=100):
-    """Evaluate fitted model using a set of metrics.
+  """Evaluate fitted model using a set of metrics.
 
-    Parameters
-    ----------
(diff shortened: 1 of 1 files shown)


edward/criticisms.py
```diff
@@ -83,16 +83,16 @@ def evaluate(metrics, model, variational, data, y_true=None, n_samples=100):
     elif metric == 'squared_hinge':
       evaluations += [sess.run(squared_hinge(y_true, y_pred))]
     elif metric == 'mse' or metric == 'MSE' or \
-       metric == 'mean_squared_error':
+            metric == 'mean_squared_error':
```
dustinvtran: what's the pep8/tensorflow convention for these line breaks? is it 8 spaces, 2 spaces, 4 spaces, or should they be aligned with the `metric ==...` line above?

PR #225, opened 2016-08-16
Feature/pep8 examples

akucukelbir:
#### Summary:

+ converts all files under `/examples` to pass PEP8

#### Intended Effect:

See above.

#### How to Verify:

Look at code. Run tests.

#### Side Effects:

None.

#### Documentation:

None.

#### Reviewer Suggestions:

@dustinvtran

### DIFF
--- a/examples/convolutional_vae_util.py
+++ b/examples/convolutional_vae_util.py
@@ -21,6 +21,7 @@ from prettytensor.pretty_tensor_class import PAD_SAME
 from prettytensor.pretty_tensor_class import Phase
 from prettytensor.pretty_tensor_class import PROVIDED
 
+
 # pylint: disable=redefined-outer-name,invalid-name
 @prettytensor.Register(
     assign_defaults=('activation_fn', 'l2loss', 'stddev', 'batch_normalize'))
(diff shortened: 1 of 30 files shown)


examples/convolutional_vae_util.py
```diff
@@ -21,6 +21,7 @@
 from prettytensor.pretty_tensor_class import Phase
 from prettytensor.pretty_tensor_class import PROVIDED
 
+
```
dustinvtran: This file is copy-and-pasted from [here](https://github.com/ikostrikov/TensorFlow-VAE-GAN-DRAW/blob/master/deconv.py). Since we modify it, maybe it makes sense to add to the top of this file something like:
```
# This file is taken from 
# https://github.com/ikostrikov/TensorFlow-VAE-GAN-DRAW/blob/master/deconv.py
# We only modify the formatting to pass PEP 8 standards.
#
# Copyright 2015 Google Inc. All Rights Reserved.
....
```

akucukelbir: awesome. go for it.

akucukelbir: make sure the comment passes PEP8 ;)

PR #246, opened 2016-09-21
Fix: Adagrad was used instead of Momentum.

bayerj:
When the user requested the use of the momentum optimizer by passing of
a string, Adagrad was used instead.


dustinvtran:
Thanks Justin!

PR #256, opened 2016-09-26
add binomial, multinomial; update random_variable methods; add params

dustinvtran:
This adds three features for random variables in Edward.

+ `Binomial` and `Multinomial` random variables are added. (They are available in `tf.contrib.distributions` in TensorFlow v0.10.)
+ Parameters of every random variable can now be accessed from the `RandomVariable` objects. For example,
```python
from edward.models import Bernoulli

x = Bernoulli(p=0.5)
print(x.p)
## Tensor("RandomVariable/Bernoulli/p/control_dependency:0", shape=(), dtype=float32)
```
+ API is updated for multivariate normals and `RandomVariable`. It follows API changes to `tf.contrib.distributions` released in v0.10. The former now always uses `mu` and `sigma` for its parameters. The latter no longer has a `parameters()` method.

### DIFF
--- a/edward/models/random_variable.py
+++ b/edward/models/random_variable.py
@@ -43,28 +43,12 @@ class RandomVariable(object):
       self._value = self._dist.sample()
 
   @property
-  def distribution(self):
-    return self._dist
-
-  @property
   def name(self):
     return self._name
 
   @property
-  def dtype(self):
-    return self.distribution.dtype
-
-  @property
-  def parameters(self):
-    return self.distribution.parameters
-
-  @property
-  def is_continuous(self):
-    return self.distribution.is_continuous
-
-  @property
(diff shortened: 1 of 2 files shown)


dustinvtran:
Actually I was wrong about adding binomial and multinomial support. It turns out that their `sample_n()` method is not implemented yet. [Therefore I had to remove them.](https://github.com/blei-lab/edward/commit/20311014daa0abd9a02aeaf9d201c5eb5b380f38)

dustinvtran:
The error says it all: methods such as `log_prob` exist for it, but to instantiate it you need to work around `sample` not being implemented.

jhmarcus:
@dustinvtran ah ok thanks!

PR #260, opened 2016-09-28
tease out variable inits from inference.initialize()

dustinvtran:
One-time TensorFlow calls are taken out of `inference.initialize()`. `inference.initialize()` is now only concerned with initialization of the inference algorithm itself; see [docstrings](https://github.com/blei-lab/edward/blob/b0494b0759a186087f75dba5d2b3fb8a96b39f19/edward/inferences/inference.py#L121).

Here are two applications you can now do with this pull request.

You run variational inference to learn an approximate posterior.  You then use the approximate posterior as a proposal distribution; you run Metropolis-Hastings.
```python
inference_proposal = ed.KLqp({z: qz}, {x: x_data})
inference_proposal.run()

empirical_z_params = tf.Variable(tf.zeros(1e4))
empirical_z = Empirical(params=empirical_z_params)

inference = ed.MetropolisHastings(
    {z: empirical_z}, 
    proposal_vars={z: qz}, 
    data={x: x_data})
inference.run(variables=[empirical_z_params])
```

You roll your own EM algorithm, using MAP for some latent variables and VI for other latent variables.
```python
inference_e = ed.KLqp({z: qz}, {x: x_data})
inference_m = ed.MAP({beta: qbeta}, {x: x_data})

inference_e.initialize(n_iter=1e4)
inference_m.initialize(n_iter=1e4)

init.run()

for _ in range(inference_e.n_iter):
  inference_e.update()
  inference_m.update()
```
(with the caveat that i haven't thought about how to handle conditionals in such an algorithm here, where we are e.g., doing vi on `p(z | x, beta)` and not `p(z | x)`. my guess is that this should be implicit somehow.)

### DIFF
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -117,27 +117,69 @@ class Inference(object):
           # If key is a placeholder, then don't modify its fed value.
           self.data[key] = value
 
-  def run(self, *args, **kwargs):
+  def run(self, logdir=None, variables=None, use_coordinator=True,
+          *args, **kwargs):
     """A simple wrapper to run inference.
 
-    1. Initialize via ``initialize``.
-    2. Run ``update`` for ``self.n_iter`` iterations.
-    3. While running, ``print_progress``.
-    4. Finalize via ``finalize``.
+    1. Initialize algorithm via ``initialize``.
+    2. (Optional) Build a ``tf.train.SummaryWriter`` for TensorBoard.
+    3. (Optional) Initialize TensorFlow variables.
+    4. (Optional) Start queue runners.
+    5. Run ``update`` for ``self.n_iter`` iterations.
+    6. While running, ``print_progress``.
+    7. Finalize algorithm via ``finalize``.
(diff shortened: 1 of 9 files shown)


dustinvtran:
There's an open question of whether to rename `initialize()` to something like `compile()`, `build()`, or `set_hyperparameters()`. The con with `initialize()` is that `initialize()` can be confused by users to also assume it initializes TensorFlow variables too. I don't like `build()` because that implies it only builds things in the computational graph; the method does more than this, also setting hyperparameters. In the same fashion, I don't like `set_hyperparameters()`.

`compile()` is arguably closer and is what Keras uses. My only complaint is that it can get confused with compilation of something lower level, whether it be the TensorFlow graph (whatever that means, which could certainly be interpreted by a newcomer to TensorFlow), or Python code. Also, it's strange to have `compile()` while still having `finalize()` as a method which runs at the end of the iterative algorithm. Arguably `initialize()` makes more sense here.

Issue #35, opened 2016-03-07
tutorials


dustinvtran:
[GPy has incredible tutorial notebooks](http://nbviewer.jupyter.org/github/SheffieldML/notebook/blob/master/GPy/index.ipynb). They teach basic concepts related to GPs and use GPy as a basis for explaining these concepts. I think it's a great idea, and I've personally found them useful as I dug though the GP literature. 

I think we can do something similar. Each notebook instructively explains a concept. The code is similar to the end-to-end examples currently in the repo, but with more words and explanation about the various options that are used. Here are some examples of concepts to teach:
- Black box variational inference
- Mixture modeling with stochastic variational inference
- Posterior predictive check

dustinvtran:
I am not sure if we wrote this down already, but I am doing it in case we didn't here as it's probably the most relevant location. Following discussion with Dave and Alp, it makes sense to organize the more in-depth tutorials via:

modeling
+ mixture of gaussian's
+ hierarchical linear model
+ latent Dirichlet allocation
+ variational auto-encoder

inference
+ coordinate ascent VI
+ SVI with local vs global latent variable updates
+ BBVI

criticism
+ PPC's
+ leave-one-out CV

Also following discussion with the Google Brain folks, Rif suggests we (as in Bayesflow + Edward) should focus very much on the following trinity of model examples: mixture of Gaussian's, a hierarchical linear model, and a variational auto-encoder.

dustinvtran:
+ How do do greedy layer-wise training in Edward, where one layer is learned with the others held fixed? As another example, how do we separate learning rates for the generator network and inference network in sigmoid belief nets? Define variable scopes. Then define separate optimizers for each variable scope. Let's have a inference tutorial demonstrating this.

dustinvtran:
+ (in inference) model-specific inference algorithm such as poisson factorization, explaining poisson factorization and its inference, its implementation, and a demo of its use

dustinvtran:
+ (in inference) inference networks
+ (in inference) variational EM

dustinvtran:
A few tutorials are up and online now! I think we set up an excellent foundation for both organizing the structure behind the tutorials and also building the momentum in writing them.

The scaffolding of documentation is currently

1. API (`*.rst`)
2. Tutorials (`*.tex`/`*.html`)
3. `examples/` directory (`*.py`)

[...]

PR #261, opened 2016-09-28
[MRG] Getting-Started Notebook

bhargavvader:
@dustinvtran, this is with respect to #35.

This is just the getting started - "First Edward Program" done in a notebook format.
It combines both the source code and the webpage's material.

Do let me know if this is fine, I'll try and do the same for all the other tutorials and example programs.


bhargavvader:
@dustinvtran, what's up with the `Coverage` check? Why is it still pending?

Would also like a review of this when you can, so I can start working on the other tutorial notebooks. :)

dustinvtran:
Looks great! Do you have a suggestion of where to link to these? My guess is that if there's an iPython notebook for each example, we can add a sentence which links to this `docs/notebooks` directory in the main Tutorials page.

bhargavvader:
Yeah, that sounds good! 
I also haven't verified this yet - but do all the `examples` feature in the tutorials section?
Or are there some python files in `example` which do not have tutorials written up?

dustinvtran:
Yeah quite a few python files in the `examples/` directory do not have a tutorial. For now I guess it makes sense to only feature iPython notebooks for examples that have tutorials. (Examples that don't have tutorials are more raw/bleeding edge and are likely to change.)

bhargavvader:
Okay, cool. :)
Feel free to merge this now if you think it is fine. 

I'll get around to doing the same for the other tutorials as well, and maybe make some 'temporary' notebooks for the example files which don't have tutorials written up, so it can maybe serve as a way to write the tutorials for those, and serve as a prototype (and also helps me in understanding edward better when I do this).

PR #273, opened 2016-10-02
Include `feed_dict` kwarg in `Inference.run()`

pdeford:
Allow user to include keyword argument `feed_dict` when calling `run` on `Inference` objects.

Covers cases where including mappings to placeholders in `data` fails.

### DIFF
--- a/docs/tex/tutorials/decoder.tex
+++ b/docs/tex/tutorials/decoder.tex
@@ -54,6 +54,8 @@ from 0 and 1.
 
 An example script using this model can found
 \href{https://github.com/blei-lab/edward/blob/master/examples/vae.py}
+{here}. An example with a convolutional architecture can be found
+\href{https://github.com/blei-lab/edward/blob/master/examples/convolutional_vae_prettytensor.py}
 {here}.
 %We experiment with this model in the
 %\href{/tutorials/variational-autoencoder}{variational auto-encoder} tutorial.
--- a/docs/tex/tutorials/inference-networks.tex
+++ b/docs/tex/tutorials/inference-networks.tex
@@ -95,6 +95,9 @@ gradients of the variational objective.
 An example script using this variational model can found
 \href{https://github.com/blei-lab/edward/blob/master/examples/vae.py}
 {here}.
+An example with a convolutional architecture can be found
+\href{https://github.com/blei-lab/edward/blob/master/examples/convolutional_vae_prettytensor.py}
+{here}.
 %We experiment with this model in the
 %\href{/tutorials/variational-autoencoder}{variational auto-encoder} tutorial.
 
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -164,7 +164,13 @@ class Inference(object):
     else:
       init = tf.initialize_variables(variables)
 
-    init.run()
+    # Feed placeholders in case initialization depends on them.
+    feed_dict = {}
+    for key, value in six.iteritems(self.data):
+      if isinstance(key, tf.Tensor):
+        feed_dict[key] = value
+
+    init.run(feed_dict)
 
     if use_coordinator:
       # Start input enqueue threads.
--- a/edward/inferences/klpq.py
+++ b/edward/inferences/klpq.py
@@ -77,15 +77,15 @@ class KLpq(VariationalInference):
         q_log_prob[s] += tf.reduce_sum(
             qz.log_prob(tf.stop_gradient(z_sample[z])))
 
-      # Form dictionary in order to replace conditioning on prior or
-      # observed variable with conditioning on posterior sample or
-      # observed data.
-      dict_swap = z_sample
-      for x, obs in six.iteritems(self.data):
-        if isinstance(x, RandomVariable):
-          dict_swap[x] = obs
-
       if self.model_wrapper is None:
+        # Form dictionary in order to replace conditioning on prior or
+        # observed variable with conditioning on posterior sample or
+        # observed data.
+        dict_swap = z_sample
+        for x, obs in six.iteritems(self.data):
+          if isinstance(x, RandomVariable):
+            dict_swap[x] = obs
+
         for z in six.iterkeys(self.latent_vars):
           z_copy = copy(z, dict_swap, scope='inference_' + str(s))
           p_log_prob[s] += tf.reduce_sum(z_copy.log_prob(z_sample[z]))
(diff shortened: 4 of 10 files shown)


dustinvtran:
Did you mean to pass `feed_dict` as an argument into `self.update()`? If so, I guess a useful question is to ask when a user would apply this.

If the placeholder is fed a value that is constant throughout inference, you can pass it into data. For example, in Bayesian linear regression,
```python
X = ed.placeholder(tf.float32, [N, D])
w = Normal(mu=tf.zeros(D), sigma=tf.ones(D))
b = Normal(mu=tf.zeros(1), sigma=tf.ones(1))
y = Normal(mu=ed.dot(X, w) + b, sigma=tf.ones(N))

...

data = {y: y_data, X: x_data}
inference = ed.Inference(..., data)
```
For manually fed values, you can pass it during each iteration of the algorithm with `inference.update(feed_dict=...)`.
```python
inference = ed.Inference(..., data)
inference.initialize()

init.run()

for _ in range(inference.n_iter):
  info_dict = inference.update(feed_dict=...)
  inference.print_progress(info_dict)
```
I guess there's a middleground between those two–when the user creates `feed_dict` with a Python generator?

pdeford:
Hi Dustin,

It might be helpful if I explain what brought me to this use case.

I was following along the [Bayesian Neural Net example](https://github.com/blei-lab/edward/blob/master/examples/bayesian_nn.py), and wanted to check my understanding by adjust the model to fit a dataset that I have on hand. The dataset was approximately 20Gb in size, but Tensors can only accomodate 2Gb. I therefore used a placeholder.

I tried including the mapping of the placeholder to my data in `data` and feed that into the `Inference` object as suggested:
> ```data = {y: y_data, X: x_data}
inference = ed.Inference(..., data)```

However, when I called `inference.run(...)` the mapping wouldn't take, and I would get errors referring to the placeholder not being filled in. As far as I could tell, this was the result of `init.run()` being called without any arguments within the `Inference` object. By letting it accommodate the `feed_dict`, I was able to manually feed in my large data set at run time.

I will try using the `update` method to see how that compares, and will get back to you. 

Thanks for your response!

dustinvtran:
Have you tried it on the latest development version?
```python
pip install -e "git+https://github.com/blei-lab/edward.git#egg=edward"
```
A recent pull request (#253) may have fixed this by having a default `feed_dict` for feeding placeholders passed into the data dictionary. [[source here]](https://github.com/blei-lab/edward/blob/41dd9a7a6e613095eca913694b828c088ebc2317/edward/inferences/variational_inference.py#L203)

pdeford:
I have installed the latest development version as recommended. It was my experience that my placeholder values were still not filled in. 

Additionally, I tried using the `update` method. Prior to calling `update`, it was necessary to call the `initialize` method. It was at this stage that the error was raised regarding the placeholder not being filled in.

dustinvtran:
I see. So in your example, you define TensorFlow variables whose initialization depends on the placeholders? E.g., something like
```python
>>> x = ed.placeholder(tf.float32, [5])
>>> y = tf.Variable(x)
>>> init = tf.initialize_all_variables()
>>> sess = ed.get_session()
>>> sess.run(init)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/dvt/Envs/venv-private/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 717, in run
    run_metadata_ptr)
  File "/Users/dvt/Envs/venv-private/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 915, in _run
    feed_dict_string, options, run_metadata)
  File "/Users/dvt/Envs/venv-private/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 965, in _do_run
    target_list, options, run_metadata)
  File "/Users/dvt/Envs/venv-private/lib/python2.7/site-packages/tensorflow/python/client/session.py", line 985, in _do_call
    raise type(e)(node_def, op, message)
tensorflow.python.framework.errors.InvalidArgumentError: You must feed a value for placeholder tensor 'Placeholder' with dtype float and shape [5]
	 [[Node: Placeholder = Placeholder[dtype=DT_FLOAT, shape=[5], _device="/job:localhost/replica:0/task:0/cpu:0"]()]]
```
(If possible, can you provide the script?)

pdeford:
That is correct. The sample you provide succinctly sums up the behavior I have seen. For your reference, [here is the script](https://github.com/pdeford/temporary/blob/master/neural_net_test.py) that I am using. It relies on `PyTables`, and the data comes from an HDF5 file with data sets `X` (2d array) and `Y` (1d array) under the `root` group.

Thanks for taking a look at this.

dustinvtran:
I don't think you need a `tf.Variable()` in this setting. Unless I'm mistaken it seems like you could just do
```python
...
x_place = ed.placeholder(tf.float32, shape=x_train.shape)
y = Normal(mu=neural_network(x_place, ...
...
```
and all would be valid. (Note that the above uses `ed.placeholder()` instead of `tf.placeholder()`; `ed.placeholder()` is just a simple wrapper that also adds TensorFlow placeholders to a TensorFlow collection; this is necessary downstream during inference for various technical reasons.)

I think your point does hold in general though. And in such a setting, just like with `inference.update()`, it can be fixed by letting the user pass it into the data dictionary, and have `init.run()()` automatically generate the `feed_dict` from this data dictionary. E.g.,
```python
    for key, value in six.iteritems(self.data):
      if not isinstance(key, RandomVariable) and not isinstance(key, str):
        feed_dict[key] = value

    init.run(feed_dict)
```

dustinvtran:
Here's my suggested fix explicitly: https://github.com/blei-lab/edward/compare/feature/init-feed-dict

Let me know if you think this solves your problem. If so, I'd like you to get contributor credit. Feel free to make this change, submit another PR, and I'll merge it.

pdeford:
That fixes it perfectly! Also, you were absolutely correct regarding `x_place` directly without a `tf.Variable`, and I took your advice to use the `ed.placeholder` wrapper.

Thanks so much for all of your help and attention with this. I am really looking forward to using `edward` more.

pdeford:
I have reverted the changes I originally proposed, and added the changes you suggested, on the latest version of `master`.

dustinvtran:
Cool! can you rebase the commits to be a single commit? e.g., `git rebase -i HEAD~4`, then collapse all commits as a single one in the text editor that appears

pdeford:
After struggling with my commits, I forked a fresh copy and made a new pull request with just the one commit. Sorry for the mess!

PR #288, opened 2016-10-06
feed placeholders if variable initialization depends on them (#273)

pdeford:
A fresh commit allowing placeholders to be fed if variable initialization depends on them, for instance of the data exceeds the 2Gb limits for Tensors


dustinvtran:
lgtm

PR #294, opened 2016-10-07
Cox model example

diengadji:
This implements a Cox model on toy data. Uses multivariate_normal_cholesky for the log prior to avoid ill-conditioned kernel matrix.


examples/cox.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python
+"""Cox process model using mean-field variational inference.
+   Ref: Miller et al, 2014.
```
dustinvtran: does andrew miller et al's work use a form of black box variational inference? otherwise it may be better to cite a more classical cox process paper.

examples/cox.py
```diff
@@ -0,0 +1,158 @@
...
+    zs: latent variables
+    	size NxV
+    Returns: log joint log p(xs, zs), scalar	
+    """
+    x = xs['x']
+    z = zs['z']
+    z = tf.reshape(z, [self.N, self.V])
+
+    log_prior = multivariate_normal_cholesky.logpdf(
+			z, tf.zeros([self.N, self.V]), self.kernel(x))
```
dustinvtran: `multivariate_normal_cholesky` takes the cholesky factor as input and not the covariance matrix.

examples/cox.py
```diff
@@ -0,0 +1,158 @@
...
+from __future__ import print_function
+
+import edward as ed
+import numpy as np
+import tensorflow as tf
+
+from edward.models import Normal
+from edward.stats import poisson, multivariate_normal_cholesky
+from edward.util import multivariate_rbf
+from scipy.stats import multivariate_normal as mvn, poisson as poi
```
dustinvtran: following pep8 style, i recommend using their full names and not abbreviating

examples/cox.py
diengadji: I added the origin paper from Cox himself...Cox, 1955. I do think we need to keep Miller et al as reference since that might be quicker to look at than the origin Cox work for someone who wants to quickly work out the details of the model. In Miller et al they do Elliptical slice sampling...Adding their reference was to give a reference to the model not the algorithm...

examples/cox.py
diengadji: I see. I already have the name poisson from edward.stats. But yes this can be updated to reflect your comment.

examples/cox.py
diengadji: great catch! solved this with a check when computing the kernel matrix...the solution i found was to add a very small number to the diagonal to ensure positive definiteness. Im open to other suggestions if you know some other solutions.

examples/cox.py
dustinvtran: oh i see. yeah keeping it makes sense!

examples/cox.py
dustinvtran: one suggestion is do `from scipy import stats` and then call `stats.multivariate_normal` and `stats.poisson`

examples/cox.py
```diff
@@ -0,0 +1,164 @@
...
+    x = xs['x']
+    z = zs['z']
+    z = tf.reshape(z, [self.N, self.V])
+
+    chol = tf.cholesky(self.kernel(x))
+
+    log_prior = multivariate_normal_cholesky.logpdf(
+      z, tf.zeros([self.N, self.V]), chol)
+
+    log_lik = poisson.logpmf(x, mu=self.inverse_link(z))
```
dustinvtran: just noticed this—does `mu` work? iirc the parameter argument for poisson is [`lam`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/poisson.py)

examples/cox.py
dustinvtran: that makes sense. its awesome that the dimensions line up so that you can apply `tf.cholesky()` directly on `self.kernel(x)` without any reshaping.

dustinvtran:
by the way, i think an implementation with Edward's language would look like
```python
x_ph = ed.placeholder(tf.float32, [N, V])  # (N, V)
K = kernel(x)  # (N, V, V)
z = MultivariateNormalCholesky(mu=tf.zeros([N, V]), chol=tf.cholesky(K))  # (N, V)
x = Poisson(lam=inverse_link(z))  # (N, V)

...

data = {x: df, x_ph: df}
inference = ed.MFVI({z: qz}, data)
```

examples/cox.py
diengadji: yeah that's a good idea...will do this in the next commit. thx

examples/cox.py
diengadji: yeah the broadcasting is handled nicely!

examples/cox.py
diengadji: in edward the poisson distribution takes \mu as the name for lambda in poisson...this works! Although i have to admit that my first try was to use lam but had to correct that and take \mu after checking the api

diengadji:
I checked at the pep fails and it looks like these are indentation and spacing that pep8 doesn't like...was this enforced in edward? I will try to fix them in the next commit...

examples/cox.py
dustinvtran: oh wow that's bad on my part. i'll submit a separate PR to fix that issue.

dustinvtran: fixed here (#298)

dustinvtran:
Running [`pep8`](https://pypi.python.org/pypi/pep8) locally on your files might be helpful.

dustinvtran:
With @jmxpearson's pull request enabling custom inits to `RandomVariable`, we can now work with the Cox process in Edward's native language. To do this, my proposal above works with the addition of one argument to `Poisson`: `value=tf.zeros_like(z)`.
A similar solution is used in [`latent_space_model.py`](https://github.com/blei-lab/edward/blob/master/examples/latent_space_model.py#L39).

diengadji:
sounds good

dustinvtran:
hi @diengadji: i revised the script. the loss seems to be converging. do you have any python plots we can use to check the fit? it would be really cool to apply this to the NBA data set we did for the chi vi paper.

diengadji:
sure will run it with one of the datasets. although the kernel here will be on the court and not on the shot data. i will revise the script accordingly.

PR #301, opened 2016-10-10
Cleaned up getting-started notebook

bhargavvader:
Made some changes to the getting-started notebook - particularly, it mentioned 1000 iterations while only running 500, there was an extra empty cell, and the final plot for after inference was run was not there. 

It also includes the neater print statement which shows the loss.


bhargavvader:
This fails a python 3.4 check for an import which I think is unrelated.

dustinvtran:
Thanks. The loss function seems quite stochastic. Could you run it for 1000 iterations so that the values look more like they're converging? (The plot also has iteration 1000 in the title anyways.)

re:travis. Yeah it seems quite finicky after #298 was merged. I'll look into it.

bhargavvader:
Yup, made it 1000.

PR #302, opened 2016-10-10
Optional print in Inference run.

bhargavvader:
I initially wanted to make this change when the print_progress output was rather messy - but post PR #259 things are a lot better.

But at any rate I felt the user should always be able to toggle whether he wants to see results or not - this will also help in making notebooks if the log is too long or unnecessary. 

Just included an option to make `print_progress` false, where default is true.

### DIFF
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -118,7 +118,7 @@ class Inference(object):
           # If key is a placeholder, then don't modify its fed value.
           self.data[key] = value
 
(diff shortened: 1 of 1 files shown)


bhargavvader:
Again, errors seem to be unrelated import issues...

dustinvtran:
You can suppress it by specifying `n_print=0`. :)

bhargavvader:
Uh oh, never noticed that.

PR #303, opened 2016-10-12
Bayesian Linear regression Notebook

bhargavvader:
The current notebook regression seemed a little bland, with no texts or comments about what is going on... I added texts and comments from the tutorial and the example file, restructured some of the notebook rows, and just added a visualisation method so it looks neater.

Only cosmetic changes.


dustinvtran:
Looks great! Only a minor comment, which is that it needs LaTeX markup (i think ipynb supports that). Thanks for adding the visualization. This is helpful for understanding the model fit. For consistency, could you also add the same added text/visuals into the `tex/...` tutorial? And the associated Python script.

bhargavvader:
The text I added was only from the `tex` tutorial and the example script - maybe some restructuring of the sentence to make it sound better, but no new text (or ideas). I don't think there is anything to add to the tex or the python script!

I'll add the LaTeX markup to the notebook.

bhargavvader:
Added the LaTeX markup.

dustinvtran:
Awesome!

The README in the [`docs/`](https://github.com/blei-lab/edward/tree/master/docs) describes how to view it locally. It doesn't require a LaTeX install, as it uses pandoc to convert it to HTML, which you can then open.

bhargavvader:
I'm a bit confused about what to do after running `./deploy.sh` - I get a page build warning, (obviously) because the edwardlib.org is taken. But I just want to run this locally - any instructions on how to do that?

dustinvtran:
After you run `./compile.sh`, you can simply open up the HTML files that are created. The webpages should all be rendered correctly as they are static. (For the links to work properly, I recommend using a local HTTP server, such as from [node js](https://www.npmjs.com/package/http-server).)

bhargavvader:
Sorry for the delay with this - was struggling with editing the webpages, but figured it out, will get to this soon. :)

dustinvtran:
sure, feel free to take your time.

bhargavvader:
@dustinvtran , I changed parts of the notebook so it is upto date with the changes caused in #344  (`scipy.stats`, imports and `KLpq`).

Also, I've failed rather miserably in editing the webpages to include the images (i'm really terrible at any web stuff). Could I (or anyone else) open another PR to take a stab at that? You could merge this one if you think it's fine.

dustinvtran:
Sure, we can open another PR to take a stab at that.

All looks good for this PR. Two notes:

> In this case $\text{MFVI}$ defaults to minimizing the ...

In this case we minimize the ...

> #likelihood of the data

`# likelihood of the data`

bhargavvader:
Yup, fixed the typos. :)

dustinvtran:
thanks!

PR #316, opened 2016-10-29
stable gaussian process classification with global kernel

diengadji:
This provides a more stable GPC code...The changes are the following:

1. added a kernel() method to the model wrapper that uses the existing multivariate.rbf() method to compute the kernel matrix for the whole data

2. the kernel has been stabilized by adding a negligible number in its diagonal...

Note: this uses the benchmark crabs dataset from UCI. The data is shuffled...


dustinvtran:
cool! could we have a version of this that uses the modeling language? currently, the `examples/` directory is trying to consist of scripts that only use the modeling language. for those using model wrappers, we prepend the filename, e.g., with `tf_` to get `tf_gp_classification.py`.

dustinvtran:
i added a [`gp_classification.py`](https://github.com/blei-lab/edward/blob/5afba38eef7a6ffff1824f6c373d3499d5e416e5/examples/gp_classification.py) file, which uses Edward's language. this could help you get started with my comment above.

diengadji:
@dustinvtran : yes this will be helpful. Thx! I was having "cannot copy z" errors with Edward language...So I will submit the initial code I have that do not use Edward language as tf_gpc.py and add yours as gpc.py into one PR. Will get back to you this weekend.

diengadji:
@dustinvtran : this is ready to merge...

edward/util/tensorflow.py
```diff
@@ -310,6 +310,44 @@ def multivariate_rbf(x, y=0.0, sigma=1.0, l=1.0):
       tf.exp(-1.0 / (2.0 * tf.pow(l, 2.0)) * tf.reduce_sum(tf.pow(x - y, 2.0)))
 
 
+def multivariate_rbf_kernel(x, sigma=1.0, l=1.0):
```
dustinvtran: when a function is added to the codebase in `util`, it needs a unit test.

edward/util/tensorflow.py
```diff
@@ -310,6 +310,44 @@ def multivariate_rbf(x, y=0.0, sigma=1.0, l=1.0):
...
+  """
+  computes the rbf kernel for the whole data x
+  Args:
```
dustinvtran: following our convention, can we convert this docstring style to be NumPy?

edward/util/tensorflow.py
```diff
@@ -310,6 +310,44 @@ def multivariate_rbf(x, y=0.0, sigma=1.0, l=1.0):
...
+  sigma:
+    standard deviation
+  l:
+    scaling parameter
+  Returns:
+     mat: a matrix of size NxN
+  """
+  N = x.get_shape()[0]
+  mat = []
+  for i in range(N):
```
dustinvtran: is there any way we can convert the multivariate rbf to be vectorized in some way, and not have to loop over every entry in the matrix? i've found this to be the computational bottleneck in our GP experiments.

looking at GPflow's [`kernels.py`](https://github.com/GPflow/GPflow/blob/master/GPflow/kernels.py), i couldn't find out how (or if) they vectorize this operation.

examples/gpc.py
```diff
@@ -0,0 +1,40 @@
...
+# DATA
+df = np.loadtxt('data/crabs_train.txt', dtype='float32', delimiter=',')
+df[df[:, 0] == -1, 0] = 0
+N = len(df)
+D = df.shape[1] - 1
+permutation = np.random.choice(range(N), N, replace=False)
+X_train = df[:, 1:][permutation]
+y_train = df[:, 0][permutation]
+
+print("pre-computing the kernel matrix...")
```
dustinvtran: Can you elaborate on what you mean by "pre-computing the kernel matrix"?

examples/tf_gpc.py
```diff
@@ -0,0 +1,75 @@
...
+ed.set_seed(54)
+
+df = np.loadtxt('data/crabs_train.txt', dtype='float32', delimiter=',')
+N = len(df)
+permutation = np.random.choice(range(N), N, replace=False)
+x = df[:, 1:][permutation]
+y = df[:, 0][permutation]
+
+print("computing the kernel matrix...")
+K = multivariate_rbf_kernel(
```
dustinvtran: if we're building this as part of the model, it would be nice to prevent global scoping for classes and have `K` be an argument to `GaussianProcess`. `self.N` and `self.n_vars` for example could also be inferred from the shape of `K`.

dustinvtran:
Thanks for the update! It looks great. Comments above.

examples/gpc.py
diengadji: i meant compute K externally and not inside the model definition...

examples/tf_gpc.py
diengadji: yes passing K as an argument is also an option. will add that.

edward/util/tensorflow.py
diengadji: sure.

edward/util/tensorflow.py
diengadji: multivariate_rbf_kernel() might be needed for any model with a RBF kernel such as GP, Cox. 

in the long term i think we might want to implement other kernels and not just RBF. Probably not in util but in a kernel.py file as is done in GPflow. Then in the example scripts we can simply do:

from ed.kernel import rbf, linear
K = rbf(x)
K_lin = linear(x)

diengadji: i will move it to the example scripts for now.

examples/gpc.py
dustinvtran: got it. i guess that makes sense for the model wrapper. everything is already "pre-computed" in the native language.

diengadji:
@dustinvtran : i am not sure why the python3.4 check is failing. this pr is ready once that is sorted out.

dawenl:
That's a legacy issue which has already been fixed in #324.

diengadji:
@dawenl : that's great! thanks

dustinvtran:
cool! since there's already a `gp_classification.py`, can you replace that script with these lines of code? (preferably keeping the filename `gp_classification.py`)

dustinvtran:
Improved via #596.

PR #318, opened 2016-11-05
Fixes an issue where a new Div node is created every update

matthewdhoffman:
The division in the line
```
_, accept_rate = sess.run([self.train, self.n_accept / self.t], feed_dict)
```
creates a new op every time it's run, which is once per iteration. This is a huge drag on performance.


dustinvtran:
thanks matt!

PR #323, opened 2016-11-07
Add 'n' parameter to RandomVariable to access sample_n() functions

matthewdhoffman:
This is a small PR that adds a keyword argument 'n' to Edward `RandomVariables` that lets them access the `tf.contrib.distributions.Distribution.sample_n()` function. This lets us replace (IMO) hacky syntax like `ed.Normal(mu=tf.ones([10, 1])*mu_vec, sigma=1.)` with `ed.Normal(mu=mu_vec, sigma=1., n=10)`, which is (IMO) clearer, very marginally more efficient, and makes doing conjugacy algebra on the graph a little easier in some cases.


dustinvtran:
interesting proposal. i like it!

i wonder if it breaks any of the inference algorithms? my guess is no, because they all work on samples via a random variable's `value()` and not through the `sample()` method. does it affect the `log_probs`?

also: the argument `n` may have to change to a unique keyword argument that no distribution's argument uses. for example, binomial and multinomial use `n` for the number of trials.

matthewdhoffman:
Good catch on the `n` overloading. Maybe `n_samples`?

Re: `log_prob()` working, I think it should be ok, since the distributions in `tf.contrib.distribution` should be written so that `log_prob()` works with samples generated via `sample()` or `sample_n()`. But worth a quick double-check.

dustinvtran:
sure, `n_samples` is good. by the way, i recommend looking at #326 to deal with the issue with copying random variables.

matthewdhoffman:
@dustinvtran I've finally gotten back to this. I think it's doing what it should, but it'd be good to get another set of eyes on it.

We can now pass in a tuple `sample_shape`, which adds 1 or more additional outer dimensions beyond whatever is specified by the `batch_shape` and `event_shape`. This basically just wraps the `sample_shape` arg from `tf.contrib.distributions`. I had to make some small adjustments to make things play nicely with the `value` parameter. (Possibly it shouldn't be possible to specify both.)

dustinvtran:
Cool! Is there a way to see the diff? The one here says 199 files changed.

chmp:
A rebase to master should fix the number of changed files, shouldn't it?

matthewdhoffman:
I'm going to close this and open a new one based on the conjugacy PR #588.

matthewdhoffman:
Obsoleted by PR #591.

Issue #309, opened 2016-10-19
travis build failure for python 3.4


dustinvtran:
when running `edward.stats` tests, it says

```
_________ ERROR collecting tests/test-stats/test_dirichlet_entropy.py __________
ImportError while importing test module '/home/travis/build/blei-lab/edward/tests/test-stats/test_dirichlet_entropy.py'.
Original error message:
'cannot import name 'dirichlet''
```

this has been happening ever since we autogenerate `edward.stats` functions in #298

dustinvtran:
fixed in #324

PR #324, opened 2016-11-07
Fix python3 random import error

dawenl:
This PR acts as a temporary fix to #309. The problem seems to be that with Python 3 `inspect.isclass` sometimes randomly filters out some of the objects, which causes import error when unit tests try to load these objects. This is by no mean a permanent fix though.


dustinvtran:
thanks @dawenl.

PR #326, opened 2016-11-08
specify initial value in RandomVariable models

jmxpearson:
- needed for rvs like Poisson with no implemented sample()
- converts value= keyword to tensor of dtype to match distribution
- tests check that dtype of _value attribute is correct and shape
  matches batch_shape + event_shape

### DIFF
--- a/tests/test-models/test_random_variable_init_value.py
+++ b/tests/test-models/test_random_variable_init_value.py
@@ -8,7 +8,7 @@ import tensorflow as tf
 from edward.models import RandomVariable
 
 from edward.models import *
-from edward.util import get_dims
+from edward.util import get_dims, copy
 
 
 def _test(RV, value, *args, **kwargs):
@@ -18,6 +18,13 @@ def _test(RV, value, *args, **kwargs):
                     rv.get_event_shape().as_list())
   assert value_shape == 0
   assert rv.dtype == rv._value.dtype
+
+def _test_copy(RV, value, *args, **kwargs):
+  rv1 = RV(*args, value=value, **kwargs)
+  rv2 = copy(rv1)
+  value_shape1 = rv1._value.get_shape().as_list()
+  value_shape2 = rv2._value.get_shape().as_list()
+  assert value_shape1 == value_shape2
 
 
 class test_random_variable_init_value_class(tf.test.TestCase):
@@ -33,6 +40,12 @@ class test_random_variable_init_value_class(tf.test.TestCase):
           _test(Normal, 2, mu=[0.5, 0.5], sigma=1.)
           _test(Normal, 2, mu=[0.5], sigma=[1.])
           _test(Normal, [2], mu=0.5, sigma=1.)
+
+  def check_copy(self):
+    with self.test_session():
+      _test_copy(Normal, 2, mu=0.5, sigma=1.)
+      _test_copy(Normal, [2], mu=[0.5], sigma=[1.])
+      _test_copy(Poisson, 2, lam=0.5)
 
 if __name__ == '__main__':
   tf.test.main()
--- a/tests/test-models/test_random_variable_init_value.py
+++ b/tests/test-models/test_random_variable_init_value.py
@@ -16,7 +16,7 @@ def _test(RV, value, *args, **kwargs):
   value_shape = rv._value.get_shape().as_list()
   expected_shape = (rv.get_batch_shape().as_list() +
                     rv.get_event_shape().as_list())
-  assert value_shape == 0
+  assert value_shape == expected_shape 
   assert rv.dtype == rv._value.dtype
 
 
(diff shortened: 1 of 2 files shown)


dustinvtran:
great implementation. that makes sense to pop `value` out of `**kwargs` rather than hard-code the argument as the first one in `RandomVariable`. the checking and test looks great too.

can you convert your code in the `if/else` chain to use 2-space indent and not 4?

dustinvtran:
@franrruiz: if you're still playing with the poisson factorization stuff in edward, this might be of use for you.

jmxpearson:
No problem.

jmxpearson:
@dustinvtran I'm not sure what's up with coveralls, but this should be ready to go.

tests/test-models/test_random_variable_init_value.py
```diff
@@ -0,0 +1,52 @@
...
+def _test_copy(RV, value, *args, **kwargs):
+  rv1 = RV(*args, value=value, **kwargs)
+  rv2 = copy(rv1)
+  value_shape1 = rv1._value.get_shape().as_list()
+  value_shape2 = rv2._value.get_shape().as_list()
+  assert value_shape1 == value_shape2
+
+
+class test_random_variable_init_value_class(tf.test.TestCase):
+  def check_shape_and_dtype(self):
```
dustinvtran: i wasn't able to get `py.test` to run these tests until i renamed `check_*` to `test_*`

tests/test-models/test_random_variable_init_value.py
```diff
@@ -0,0 +1,52 @@
...
+    with self.test_session():
+      _test(Normal, 2, mu=0.5, sigma=1.)
+      _test(Normal, [2], mu=[0.5], sigma=[1.])
+      _test(Poisson, 2, lam=0.5)
+
+  def check_mismatch_raises(self):
+    with self.test_session():
+        with pytest.raises(ValueError):
```
dustinvtran: this produced an error for me. is there a [tensorflow way](https://www.tensorflow.org/versions/master/api_docs/python/test.html) of doing this, or which is unit testing module independent?

```python
____________________________________________________________ test_random_variable_init_value_class.test_mismatch_raises _____________________________________________________________

self = <test_random_variable_init_value.test_random_variable_init_value_class testMethod=test_mismatch_raises>

    def test_mismatch_raises(self):
      with self.test_session():
>         with pytest.raises(ValueError):
E         NameError: global name 'pytest' is not defined

tests/test-models/test_random_variable_init_value.py:40: NameError
```

tests/test-models/test_random_variable_init_value.py
dustinvtran: i also get

```python
____________________________________________________________ test_random_variable_init_value_class.test_shape_and_dtype _____________________________________________________________

self = <test_random_variable_init_value.test_random_variable_init_value_class testMethod=test_shape_and_dtype>

    def test_shape_and_dtype(self):
      with self.test_session():
>       _test(Normal, 2, mu=0.5, sigma=1.)

tests/test-models/test_random_variable_init_value.py:34:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

RV = <class 'edward.models.random_variables.Normal'>, value = 2, args = (), kwargs = {'mu': 0.5, 'sigma': 1.0}, rv = <ed.RandomVariable 'Normal/' shape=() dtype=tf.float32>
value_shape = [], expected_shape = [], @py_assert2 = 0, @py_assert1 = False, @py_format4 = '[] == 0', @py_format6 = 'assert [] == 0'

    def _test(RV, value, *args, **kwargs):
      rv = RV(*args, value=value, **kwargs)
      value_shape = rv._value.get_shape().as_list()
      expected_shape = (rv.get_batch_shape().as_list() +
                        rv.get_event_shape().as_list())
>     assert value_shape == 0
E     AssertionError: assert [] == 0

tests/test-models/test_random_variable_init_value.py:19: AssertionError```

tests/test-models/test_random_variable_init_value.py
dustinvtran: below also uses 4-space indent

tests/test-models/test_random_variable_init_value.py
jmxpearson: Apologies. I was testing these by hand; didn't think about the naming convention. The failure resulted from me deleting the wrong line. Good catch.

tests/test-models/test_random_variable_init_value.py
jmxpearson: There is. Fixed.

dustinvtran:
looks good to me. thanks!

PR #327, opened 2016-11-08
Data subsampling

### LINKED ISSUE #26: streaming data

### LINKED ISSUE #311: Reduction of losses: mean vs sum
Hey,

From what I see, the loss functions in edward are often realized as the _sum_ over data points and dimensions. See e.g. [here](https://github.com/blei-lab/edward/blob/master/edward/inferences/klqp.py#L298).

When tuning hyper parameters, the batch size is often an important one. However, in the case of summing over the data set, it is coupled to the learning rate of a optimizer, as the loss landscape is scaled.

It is therefore good practice to average the data points out, instead of summing over them.

If others think this is a good idea, I can go ahead and do the appropriate changes.

---

PS: When using models that also have spatial dimensions (e.g. `(n_time_steps, n_samples_in_batch, n_dimensons)` for time series data), there should be a convention as well. I prefer to always sum over the last dimension and take the mean over the rest.

dustinvtran:
+ issues: fix #311; fix #26 

implementation of data subsampling with proper scaling of local to global variables. also see section 4.5 of [iclr submission](http://openreview.net/forum?id=Hy6b4Pqee)

## reviewer suggestions
anyone, e.g., @bayerj, @mariru, @akucukelbir


akucukelbir:
@dustinvtran i'll review this.

dustinvtran:
i chatted with @akucukelbir in person about this. the action points are to remove `n_minibatch` for now. later we will figure out how to deal with data subsampling when the full model fits in memory. (there's this `gather` business that we have to code up eventually.)

dustinvtran:
this does not work for 
+ updates that involve global and local variables (that itself requires another feature implementation)
+ reinitialization of not only the parameters to the local factor but also anything it depends on, such as gradient history

alp will solve other things.

dustinvtran:
i rebased to solve merge conflicts from `master`

bayerj:
I thought I might chime in with two use cases that are related to these improvements. From what I have read in the code, it might already work out of the box, as the ``scale`` is basically just multiplied with the corresponding coordinate wise loss terms. But I will put them here anyway.

### Sequential models

Consider learning a sequential model (STORN, DKF, DVBF, ...). In an ideal situation, all your sequences are of the same length and fit nicely into a 3D tensor of the shape ``(N, T, D)``, where ``N`` is the number of samples, ``T`` is the sequence length and ``D`` is the data dimensionality.

In reality, all your sequences will have different lengths. To allow for parallelization, what you do is zero-pad them to the same length. Assume for now that you have a coordinate wise loss ``L``, which will also have the shape ``(N, T, D)``. Then what you can do is to multiply with a boolean mask tensor ``M`` of the shape ``(N, T, 1)`` where the elements are ``1`` iff the coordinate is not padded. To get the correct loss, you can then do ``tf.reduce_sum(L * M)``. This will waste computation, but be more efficient thanks to GPU etc.

Thinks get more complicated with ``KLqp``, as there will not only be the coordinate wise loss as above ("reconstruction loss" in VAE speak). The KL term is however of similar shape, i.e. ``(N, T, J)`` where ``J`` is the latent dimension. In this case, we can also use ``M`` to multiply the KL term with.

### Schedules

In some cases (DVBF, normalizing flows) it is beneficial to use schedules to weigh the different terms of the loss function depending on the iteration number. It would be nice if these could also be handle with a dynamic scale, i.e. the scale being a tensorflow variable.

dustinvtran:
@akucukelbir: okay. i got the `probabilistic_pca_subsampling.py` example working. aside from working on subgraphs, there were two additional tricks i employed:

+ define the full set of variational parameters; then index them accordingly as inference is run. this avoids having to reinitialize local parameters and start from scratch for every minibatch. 
+ define two inferences: a global inference over w conditional on z, and a local inference on z conditional on w. this avoids the optimizer history being the same for both global and local variables.

the doc is updated to reflect these. now waiting your greenlight. if you approve, feel free to merge.

akucukelbir:
very cool. however i'm a bit confused.

now in the ppca example we take a step in both the global variables (`w`) and the local variables (`z`) at every step. this doesn't seem to give the right answer. (try increasing the number of iterations.)

to take one global step but many local steps, i'm doing something like this:
```python
for _ in range(1000):
  x_batch, idx_batch = next_batch(M)
  info_dict = inference_w.update(feed_dict={x_ph: x_batch, idx_ph: idx_batch})
  for _ in range(25):
    inference_z.update(feed_dict={x_ph: x_batch, idx_ph: idx_batch})
  inference_w.print_progress(info_dict)
```

does this make sense? (i think it does.) 

now my question is, how is the stepsize being determined? are the two learning rates being adjusted separately via `inference_w` and `inference_z`. would it make sense to apply something like adam for `w` but rmsprop for `z`?

dustinvtran:
for the above to work you should swap the order to perform (possibly multiple) local updates before global updates.

the two learning rates are adjusted separately as they are two different inference algorithms. you could adjust and possibly reset the optimizer history for the local inference if you'd like. i haven't found it to make much of a difference though.

dustinvtran:
i have further investigated the PPCA example and tried MAP. for data subsampling and not data subsampling, the loss converges, but the parameters start to diverge in opposite directions. this tells me that it's a modeling issue.

jcapde87:
Hi Dustin, 

What is the status of the PPCA example? I have executed it and I observe that the loss converges but the `w` parameters don't, as you mentioned above.  There are couple of things that are not clear to me from this example:

- If `var_list` is the list of vars that we seek to optimise, why `inference_w` sets `var_list=qz_variables` and `inference_z` sets `var_list=qw_variables`. Shouldn't it be the other way around? 

- Regarding the scaling, I can understand that we need to scale the gradients for the global inference step `inference_w`, but which is the reason for scaling the gradients for the local inference? 

I have tried to apply both changes, but I haven't noticed any change in the convergence of `w`.

Thks

PR #337, opened 2016-11-11
RFC: adding support for tf logging in variational inf

jmxpearson:
Edward has a lot of nice wrappers to hide TF complexity, but this also makes instrumenting to check for NaNs and Infs, as well as logging, tricky. With this, I can instrument my code and use Tensorboard on the logs.

This is a small patch that allows for logging via TF summary nodes. It calls `tf.merge_all_summaries` in the appropriate place so that summary nodes are run. It also modifies variational inference to create a summary with the same frequency as progress is printed to screen.

I'm not sure I have the time to extend this to other inference methods, but the pattern should be clear.

### DIFF
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -212,7 +212,11 @@ class Inference(object):
     self.initialize(*args, **kwargs)
 
     if logdir is not None:
+      self.logging = True
       self.train_writer = tf.train.SummaryWriter(logdir, tf.get_default_graph())
+      self.summarize = tf.merge_all_summaries()
+    else:
+      self.logging = False
 
     if variables is None:
       init = tf.initialize_all_variables()
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -211,13 +211,6 @@ class Inference(object):
     """
     self.initialize(*args, **kwargs)
 
-    if logdir is not None:
-      self.logging = True
-      self.train_writer = tf.train.SummaryWriter(logdir, tf.get_default_graph())
-      self.summarize = tf.merge_all_summaries()
-    else:
-      self.logging = False
-
     if variables is None:
       init = tf.initialize_all_variables()
     else:
@@ -247,7 +240,7 @@ class Inference(object):
       self.coord.request_stop()
       self.coord.join(self.threads)
 
-  def initialize(self, n_iter=1000, n_print=None, n_minibatch=None, scale=None):
+  def initialize(self, n_iter=1000, n_print=None, n_minibatch=None, scale=None, logdir=None, debug=False):
     """Initialize inference algorithm.
 
     Parameters
@@ -268,6 +261,12 @@ class Inference(object):
       A scalar value to scale computation for any random variable that
       it is binded to. For example, this is useful for scaling
       computations with respect to local latent variables.
+    logdir : str, optional
+      Directory where event file will be written. For details,
+      see `tf.train.SummaryWriter`. Default is to write nothing.
+    debug: boolean, optional
+      If True, add checks for NaN and Inf to all computations in the graph.
+      May result in substantially slower execution times.
     """
     self.n_iter = n_iter
     if n_print is None:
@@ -311,6 +310,17 @@ class Inference(object):
       self.data = {key: value for key, value in
                    zip(six.iterkeys(dict_data), batches)}
       self.data.update(dict_rv)
+
+    if logdir is not None:
+      self.logging = True
+      self.train_writer = tf.train.SummaryWriter(logdir, tf.get_default_graph())
+      self.summarize = tf.merge_all_summaries()
+    else:
+      self.logging = False
+
+    if debug:
+      self.debug=True
+      self.op_check = tf.add_check_numerics_ops()
 
   def update(self):
     """Run one iteration of inference.
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -240,7 +240,8 @@ class Inference(object):
       self.coord.request_stop()
       self.coord.join(self.threads)
 
-  def initialize(self, n_iter=1000, n_print=None, n_minibatch=None, scale=None, logdir=None, debug=False):
+  def initialize(self, n_iter=1000, n_print=None, n_minibatch=None, scale=None,
+                 logdir=None, debug=False):
     """Initialize inference algorithm.
 
     Parameters
@@ -319,7 +320,7 @@ class Inference(object):
       self.logging = False
 
     if debug:
-      self.debug=True
+      self.debug = True
       self.op_check = tf.add_check_numerics_ops()
 
   def update(self):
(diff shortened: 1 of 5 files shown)


edward/inferences/variational_inference.py
```diff
@@ -123,6 +123,10 @@ def update(self, feed_dict=None):
 
     sess = get_session()
     _, t, loss = sess.run([self.train, self.increment_t, self.loss], feed_dict)
+    if self.logging and self.n_print != 0:
+      if t == 1 or t % self.n_print == 0:
+          summary = sess.run(self.summarize, feed_dict)
```
dustinvtran: is there a performance drop if the fetching for `self.summarize` is not in the same session run as the training stuff above?

edward/inferences/inference.py
```diff
@@ -188,7 +188,11 @@ def run(self, logdir=None, variables=None, use_coordinator=True,
     self.initialize(*args, **kwargs)
 
     if logdir is not None:
+      self.logging = True
       self.train_writer = tf.train.SummaryWriter(logdir, tf.get_default_graph())
+      self.summarize = tf.merge_all_summaries()
+    else:
+      self.logging = False
```
dustinvtran: the boolean flag makes sense. we also need to somehow enable this option when the user manually controls inference and does not call `inference.run()`. see, e.g., [`normal_idiomatic_tf.py`](https://github.com/blei-lab/edward/blob/master/examples/normal_idiomatic_tf.py), [`vae.py`](https://github.com/blei-lab/edward/blob/master/examples/vae.py).

dustinvtran:
thanks for the PR. this looks useful. extending this to other inference methods is also easy as you said. the only other one is `MonteCarlo` and it shares the same pattern.

edward/inferences/inference.py
jmxpearson: I can take a look at this.

jmxpearson:
Sounds good. I'm at a conference, but will take a closer look when I get back.

edward/inferences/variational_inference.py
dustinvtran: not much. i recommend including the `_, t, loss = ...` stuff as part of that if/else statement then.

jmxpearson: Sorry it's taken so long to get back to you on this.

I was thinking one might not want to log the summary on every iteration if `n_print != 1`. Happy to move this into line 125, though, if you think it's best.

edward/inferences/inference.py
jmxpearson: What about the idea of allowing `initialize` to take `logdir` as a parameter, setting the flag there, and moving [this](https://github.com/blei-lab/edward/pull/337#discussion-diff-87717005R128) (the actual logging) into `update`?

edward/inferences/variational_inference.py
dustinvtran: what about something like
```python
    if self.logging and self.n_print != 0:
      if t == 1 or t % self.n_print == 0:
          _, t, loss, summary = sess.run([..., self.summarize], feed_dict)
    else:
        _, t, loss = sess.run([self.train, self.increment_t, self.loss], feed_dict)
```

dustinvtran: also, no problem; feel free to take your time!

jmxpearson: So I think the issue here is that with this construction, you'd actually need
```python
if self.logging and self.n_print != 0:
      if t == 1 or t % self.n_print == 0:
          _, t, loss, summary = sess.run([self.train, self.increment_t, self.loss, self.summarize], feed_dict)
      else:
          _, t, loss, summary = sess.run([self.train, self.increment_t, self.loss], feed_dict)
else:
       _, t, loss = sess.run([self.train, self.increment_t, self.loss], feed_dict)
```
to cover the fact that you only want to summarize on the first iteration and every print iteration. My thinking was just that it seems like a lot of repetition.

jmxpearson:
Sorry for the messy history. You may want to squash before merging.

Changes:
- `logging` is now an argument to `initialize`. Check to see if you're okay with it being late, rather than early, in the argument list.
- Docstrings updated.
- I added a boolean flag, `debug`, that defaults to `False` but when turned on, has TensorFlow add `Inf` and `NaN` checks to all the floating point nodes. I've found this really helpful for debugging. It follows the same pattern as `logging` but runs on every iteration, not just on the iterations where progress is printed.

edward/models/random_variable.py
```diff
@@ -59,7 +59,8 @@ def __init__(self, *args, **kwargs):
     # need to temporarily pop value before __init__
     value = kwargs.pop('value', None)
     super(RandomVariable, self).__init__(*args, **kwargs)
-    self._kwargs['value'] = value  # reinsert (needed for copying)
+    if value is not None:
```
dustinvtran: could you describe the use case for this if clause?

dustinvtran:
all looks great. only one minor comment above.

edward/models/random_variable.py
jmxpearson: If this doesn't happen, then the subsequent copies of the node made by `copy` don't have a `value` initializer, and you run into the same problem as before.

For instance, in the original use case, because Poisson doesn't have a random number generator in TensorFlow, I need to provide this `value=` initializer because `sample` isn't defined. If I don't stick `value` back into `kwargs`, there's no initializer when the Poisson node is copied, and it throws the same error.

Make sense?

jmxpearson: Oh, and to clarify:

If one were to allow `kwargs['value'] = None`, then the subsequent copy tries to use `value=None` as an initializer, which would block the normal path (the one that calls `sample`).

dustinvtran: got it.

redst4r:
Hi,

just ran into two small issues with the logging:

1. calling `inference.run(n_iter=1000,  logdir=LOGDIR)` without any actual logging being specified by the user (no tf.summary Ops being created previously) will yield a `TypeError: Fetch argument None has invalid type <class 'NoneType'>` coming out of this call:
```
edward/edward/inferences/variational_inference.py in update(self, feed_dict)
    143     if self.logging and self.n_print != 0:
    144       if t == 1 or t % self.n_print == 0:
--> 145         summary = sess.run(self.summarize, feed_dict)
```
Basically, `self.summarize is None` (`tf.merge_all_summaries()` just returned None in inference.py line 315), because there's no Summaries to collect at that timepoint. `summarize` being None throws the mentioned Exception somewhere deep in tensorflow. 
Maybe we can just wrap it like this in `variational_inference.py`
```python
if self.summarize:
    summary = sess.run(self.summarize, feed_dict)
```

2. in inference.py, `self.train_writer.close()` is never called to close the Writer. Happend to me that some content was never flushed to disk after inference was completed. What about adding `self.train_writer.close()` to the `finalize()` method?

jmxpearson:
Good catch. That makes sense to me.

redst4r:
cool, should I open a new PR or can we somwhow put it into this one?

jmxpearson:
I suspect it will be easier to open a new PR, since there have been changes to master since this branch got merged.

redst4r:
ok, created PR #391

PR #340, opened 2016-11-13
Incorporate missing statement in getting-started

isohyt:


docs/tex/getting-started.tex
```diff
@@ -34,9 +34,11 @@ \subsubsection{Your first Edward program}
 
 \begin{lstlisting}[language=Python]
 import numpy as np
+from edward.stats import norm
```
dustinvtran: we should use scipy's as a default; i think we'll deprecate `edward.stats` soon as it doesn't serve much of a purpose. e.g.,
```
from scipy.stats import norm
```

docs/tex/getting-started.tex
```diff
@@ -34,9 +34,11 @@ \subsubsection{Your first Edward program}
...
 
 x_train = np.linspace(-3, 3, num=50)
 y_train = np.cos(x_train) + norm.rvs(0, 0.1, size=50)
+x_train = x_train.reshape((50, 1))
 \end{lstlisting}
 
 \includegraphics[width=700px]{images/getting-started-fig0.png}
```
dustinvtran: given the `mu` argument, the  `y` random variable currently has shape `(50, 1)`, which doesn't match the shape of `y_train` (`(50,)`). there are two approaches: we can flatten the `mu` argument in `y`,
```python
y = Normal(mu=tf.reshape(tf.matmul(tf.nn.tanh(tf.matmul(x, W_0) + b_0), W_1) + b_1, [-1]),
...
```
or we can also reshape `y_train`,
```python
y_train = y_train.reshape((50, 1))
```
i vote for the latter as it's easier to read, even though the full script does the former.

dustinvtran:
thanks for the PR. this is definitely worth fixing. detailed comments above

isohyt:
Oh, sorry about that. I will fix them.

PR #342, opened 2016-11-14
Coveralls fix

dawenl:


dustinvtran:
it's silly how simple of a solution that it is. then again, simple solutions are always the best solutions.

PR #348, opened 2016-11-16
use unique scopes for each inference object

### LINKED ISSUE #347: (MAP) Inference object is supposed to be stateless, but apparently not
To reproduce the problem:
```
import edward as ed
import numpy as np
import tensorflow as tf

from edward.models import Normal, PointMass

ed.set_seed(42)

# DATA
x_data = np.array([5.0] * 50, dtype=np.float32)

# MODEL: Normal-Normal with known variance
mu = Normal(mu=0.0, sigma=1.0)
x = Normal(mu=tf.ones(50) * mu, sigma=1.0)

qmu1 = PointMass(params=tf.Variable(0.0))
qmu2 = PointMass(params=tf.Variable(0.0))

# INFERENCE
data = {x: x_data}
inference = ed.MAP({mu: qmu1}, data)
inference.run()
inference = ed.MAP({mu: qmu2}, data)
inference.run()
```

The two runs will produce drastically different results. 

Only verified `ed.MAP`, maybe also affects other `Inference` subclasses.

dustinvtran:
fix #347

### DIFF
--- a/edward/inferences/sgld.py
+++ b/edward/inferences/sgld.py
@@ -97,25 +97,26 @@ class SGLD(MonteCarlo):
       Latent variable keys to samples.
     """
     if self.model_wrapper is None:
+      scope = 'inference_' + str(id(self))
       # Form dictionary in order to replace conditioning on prior or
       # observed variable with conditioning on a specific value.
       dict_swap = z_sample.copy()
       for x, qx in six.iteritems(self.data):
         if isinstance(x, RandomVariable):
           if isinstance(qx, RandomVariable):
-            qx_copy = copy(qx, scope='conditional')
+            qx_copy = copy(qx, scope=scope)
             dict_swap[x] = qx_copy.value()
           else:
             dict_swap[x] = qx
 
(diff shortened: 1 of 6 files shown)


edward/inferences/sgld.py
```diff
@@ -97,25 +97,26 @@ def log_joint(self, z_sample):
       Latent variable keys to samples.
     """
     if self.model_wrapper is None:
+      scope = 'inference_' + str(id(self))
```
dawenl: would unifying different scopes (likelihood, prior, conditional) into one "inference" scope create any collision?

dustinvtran: yeah, i'm not sure why they have different scopes in the first place. apparently i only did this for Monte Carlo methods and not any variational inference methods. i would not think not, but maybe my past self knows more.

dawenl:
Looks good to me!

dustinvtran:
cool! feel free to merge now. (use the "squash and merge" one)

PR #350, opened 2016-11-18
implement collapsed mixture models with Mixture random variable

dustinvtran:
The last tutorials written with model wrappers—the mixture of Gaussians and mixture density network—are rewritten in Edward's native language. We use the [`Mixture`](https://www.tensorflow.org/versions/r0.11/api_docs/python/contrib.distributions.html#mixture-models) random variable to collapse the mixture assignments, and which internally calculates the log-density using the log-sum-exp trick.

Neither example works. I think there's an issue with gradients not propagating correctly in the `Mixture` random variable. As inference runs, the parameters hardly change (although they do change, implying that some gradient update is occurring).

### DIFF
--- a/docs/tex/tutorials/mixture-gaussian.tex
+++ b/docs/tex/tutorials/mixture-gaussian.tex
@@ -40,24 +40,72 @@ Define the prior on each component $\mathbf{\sigma}_k\in\mathbb{R}^D$ to be
   \text{InverseGamma}(\mathbf{\sigma}_k \mid a, b).
 \end{align*}
 
-Let's build the model in Edward using TensorFlow. This simply requires
-writing down the model's log joint density,
+We build two versions of the model in Edward: one explicitly with the
+mixture assignments $c_n\in\{0,\ldots,K-1\}$ as latent variables,
+and another with them summed out.
+
+The explicit version is as follows:
+
+\begin{lstlisting}[language=Python]
+from edward.models import Categorical, Dirichlet, InverseGamma, Normal
+
+N = 500  # number of data points
+K = 2  # number of components
+D = 2  # dimensionality of data
+
+pi = Dirichlet(alpha=tf.ones(K))
+mu = Normal(mu=tf.zeros([K, D]), sigma=tf.ones([K, D]))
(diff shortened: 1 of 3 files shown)


bouliagi:
The examples fail because edward.util.copy() fails to properly replace ancestors of the Mixture random variable. The components of Mixture appear in a *list* of random variables which edward.util.copy() does not recognize as instances of the RandomVariable type.

One possible fix (which works for me) covers this case, with only a list and only one level of nesting allowed. It can be made in src/edward/edward/util/random_variables.py by adding :

    def copy_rv(value, dict_swap, scope, replace_itself, copy_q):
        if isinstance(value, RandomVariable) or \
            isinstance(value, tf.Variable) or \
             isinstance(value, tf.Tensor) or \
                isinstance(value, tf.Operation):
        value = copy(value, dict_swap, scope, replace_itself, copy_q)
    return value

And then, in def copy():

    ...
    kwargs = {}
    for key, value in six.iteritems(x._kwargs):
        if isinstance(value, list):
            kwargs[key] = [copy_rv(v, dict_swap, scope, True, copy_q) for v in value]
        else:
            kwargs[key] = copy_rv(value, dict_swap, scope, True, copy_q)
    ...
 
Without the fix, only the ancestors of the latent variables are correctly replaced, while the log_prob() of the observed variables is computed from random samples; thus the parameters converge to their priors.

dustinvtran:
that's a great catch. handling the extreme case of >1 level of nesting would be nice too, although for now we can just add as part of `copy()`'s docstring that it only handles 1 level.

could you submit a PR that makes these changes? i'm happy to merge them in so you get credit and then continue work on these mixture examples. if you have time, i also recommend adding a test to `tests/test-util/test_copy.py`, checking the copy function for lists. if you don't, i can also do it.

bouliagi:
OK, I just issued a pull request. No test though, I haven't looked at `tests/test-util/test_copy.py` to see how it is done.

dustinvtran:
test implemented in fc37c19034981c759b4a33f272752ce07aa25857

dustinvtran:
following the fix in #401, i added the collapsed mixture of gaussians and updated the tutorials. 

there's still an issue with the mixture density network which i haven't completed diagnosing. nevertheless, it can be merged in a separate PR.

PR #362, opened 2016-11-28
[WIP] Added classification notebook

bhargavvader:
Added the Gaussian process classification tutorial. This replicated the code found in the [script](https://github.com/blei-lab/edward/blob/master/examples/gp_classification.py) and the [example](http://edwardlib.org/tutorials/supervised-classification) on the website.

Thought it would be nice to visualise this as well - any ideas on how to go about doing so?

### DIFF
--- a/docs/notebooks/supervised_learning_classification.ipynb
+++ b/docs/notebooks/supervised_learning_classification.ipynb
@@ -1,0 +1,187 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


dustinvtran:
Since this example uses only one feature, perhaps you could try plotting explicitly that feature against the binary response (x vs y); then you visualize 10 function draws from the Gaussian process as 10 different curves fitting those points. In other words, you look at multiple session runs of `f.value()` (squashed between 0 and 1 with `tf.sigmoid`).

bhargavvader:
Sounds good, will have a go at it soon!

PR #364, opened 2016-11-29
Latent Networks Notebook

bhargavvader:
Notebook on the Latent Networks tutorial. It mimics the tutorial and the script. Similar to my question on the Classification notebook, I was wondering how to visualise this. The latent dimensionality here is `3`.


dustinvtran:
for this one, it makes sense to plot the data points in the latent space (`z`). this shows, for example, similarity of nodes based on their latent positions. (a good reference for this is the original paper, hoff et al. 2002).

PR #386, opened 2016-12-17
Probability Zoo

dustinvtran:
A community repository is opened for pre-trained probabilistic models. 

Part of the success of the neural net community is the ability to work with pre-trained models. The probabilistic modeling community should have this ability too.


twiecki:
Love this idea. Would it be open to stan and pymc3 models too?

dustinvtran:
sure.

we should have guidelines up soon for how to contribute a pre-trained model. my current guess is that it must include 1. the stored parameters in some (preferably easily loadable) file; and 2. a short script that loads and uses the model. an optional but nice addition would be 3. any code to reproduce the parameters.

Issue #390, opened 2017-01-04
Docstring for Models


jkerfs:
By dynamically creating the RandomVariable classes, we lose the docstrings that describe the parameters of each distribution.  This is not a major issue, but it is certainly convenient to see information about the distribution in IDEs.  Below is an example of these docstrings from pymc3.
![pymc3-docstring](https://cloud.githubusercontent.com/assets/3375795/21627197/042901a6-d1cb-11e6-9b38-6397b86a033a.png).

I thought of three ways to resolve this:
1. Inherit the docstrings from tf.contrib.distributions.

This could be achieved by replacing the following (from edward.models.randomvariables.py):
```
    class _WrapperRandomVariable(RandomVariable, _candidate):
      def __init__(self, *args, **kwargs):
        RandomVariable.__init__(self, *args, **kwargs)

    _WrapperRandomVariable.__name__ = _name
    _globals[_name] = _WrapperRandomVariable
```

with 
```
_params = {"__doc__": _candidate.__doc__ }
_globals[_name] = type(_name, (RandomVariable, _candidate), _params)
```

The downside of this is that some of the docstrings contain code samples that reference the use of `tf.contrib.distributions`.  

2.  Generate docstring from the required arguments of the distribution:
```
_n_default_args = 1 + len(_canddiate.__init__.__defaults__)
_required_variables = _candidate.__init__.__code__.co_varnames[1:-_n_default_args]
_docstring = _name + " Distribution with parameters " + ", ".join(_required_variables)
params = {"__doc__": _docstring}
```

This would produce strings like `Beta Distribution with parameters a, b` and `Normal Distribution with parameters mu, sigma`.  Unfortunately, these docstrings are not very informative, and the python code to produce them becomes verbose and muddled.

3.  Statically define the class for each distribution and provide a short custom docstring

This would require the most work, but the docstrings would be useful, the parameters for each constructor could be named (enable autocomplete and suggestions in most IDEs), and it would play nice with static documentation compilers like doxygen.  Obviously the major negative is that new distributions added to tf.contrib.distributions would have to be manually added.

dustinvtran:
Thanks for detailing the options! I agree that 3 would produce the most useful docstrings, at the cost of making it more difficult to maintain. Static classes also make it more difficult (although still possible) to support multiple tensorflow versions and not only the latest.

[...]

PR #394, opened 2017-01-09
add docstrings to random variables

jkerfs:
Resolution for Issue https://github.com/blei-lab/edward/issues/390.

Reuse the docstrings from the classes in `tensorflow.contrib.distributions` as docstrings for the `RandomVariable` classes in Edward.  Now calling `help()` on a `RandomVariable` class will show information about the distribution and the required parameters.

### DIFF
--- a/edward/models/random_variables.py
+++ b/edward/models/random_variables.py
@@ -3,7 +3,6 @@ from __future__ import division
 from __future__ import print_function
 
 import inspect
-import tensorflow as tf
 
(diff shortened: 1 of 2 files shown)


dustinvtran:
nice implementation! i hadn't thought of using `type()` here. could you add a unit test for one distribution, say, Bernoulli or Normal? for example, you can just check if the if the string `Bernoulli.__doc__` is non-empty.

dustinvtran:
thanks!

PR #401, opened 2017-01-11
changed copy() to handle list in RandomVariable _kwargs e.g. in Mixture

bouliagi:
edward.util.copy() fails to properly replace ancestors of the Mixture random variable. The components of a Mixture appear in a list of random variables which edward.util.copy() does not recognize as instances of the RandomVariable type.

This fix covers this case, but only for a list and only one level of nesting allowed. 

Also included is the addition of a var_list argument  to Map.build_loss_and_gradients() so that mixture examples can be tested with MAP inference, but this is not directly related to the copy() problem.

### DIFF
--- a/edward/util/random_variables.py
+++ b/edward/util/random_variables.py
@@ -13,6 +13,14 @@ from tensorflow.core.framework import attr_value_pb2
 from tensorflow.python.framework.ops import set_shapes_for_outputs
 from tensorflow.python.util import compat
 
+def copy_rv(value, dict_swap, scope, replace_itself, copy_q):
+    if isinstance(value, RandomVariable) or \
+        isinstance(value, tf.Variable) or \
+         isinstance(value, tf.Tensor) or \
+            isinstance(value, tf.Operation):
+        value = copy(value, dict_swap, scope, replace_itself, copy_q)
+    return value
+
 
 def copy(org_instance, dict_swap=None, scope="copied",
          replace_itself=False, copy_q=False):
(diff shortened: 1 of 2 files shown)


edward/util/random_variables.py
```diff
@@ -154,13 +162,17 @@ def copy(org_instance, dict_swap=None, scope="copied",
...
-         isinstance(value, tf.Operation):
-         value = copy(value, dict_swap, scope, True, copy_q)
+    #  if isinstance(value, RandomVariable) or \
+    #     isinstance(value, tf.Variable) or \
+    #     isinstance(value, tf.Tensor) or \
+    #     isinstance(value, tf.Operation):
+    #     value = copy(value, dict_swap, scope, True, copy_q)
+    #  kwargs[key] = value
+        if isinstance(value, list):
+            kwargs[key] = [copy_rv(v, dict_swap, scope, True, copy_q) for v in value]
```
dustinvtran: 2 space indent

edward/util/random_variables.py
```diff
@@ -154,13 +162,17 @@ def copy(org_instance, dict_swap=None, scope="copied",
 
     kwargs = {}
     for key, value in six.iteritems(rv._kwargs):
-      if isinstance(value, RandomVariable) or \
-         isinstance(value, tf.Variable) or \
-         isinstance(value, tf.Tensor) or \
-         isinstance(value, tf.Operation):
-         value = copy(value, dict_swap, scope, True, copy_q)
+    #  if isinstance(value, RandomVariable) or \
```
dustinvtran: can you delete these commented lines?

bouliagi:
Updated with your proposed corrections, I should give more attention to style next time...

dustinvtran:
perfect. i wrote a simple test that checks copying for lists. i'll add it after merging this PR.

PR #414, opened 2017-01-19
Eliminated redundant computation in repeated calls to leapfrog() function

matthewdhoffman:
In the old code, each call to `leapfrog()` computed the gradient twice with respect to latent variables that hadn't changed. This patch eliminates that duplicated computation, resulting in a near 2x speedup.

It works by adding an `n_steps` parameter to `leapfrog()` and moving the loop into the `leapfrog()` function.

### DIFF
--- a/edward/inferences/hmc.py
+++ b/edward/inferences/hmc.py
@@ -76,9 +76,9 @@ class HMC(MonteCarlo):
     # Simulate Hamiltonian dynamics.
     new_sample = old_sample
     new_r_sample = old_r_sample
-    for _ in range(self.n_steps):
-      new_sample, new_r_sample = leapfrog(new_sample, new_r_sample,
-                                          self.step_size, self._log_joint)
+    new_sample, new_r_sample = leapfrog(new_sample, new_r_sample,
+                                        self.step_size, self._log_joint,
+                                        self.n_steps)
 
     # Calculate acceptance ratio.
     ratio = tf.reduce_sum([0.5 * tf.reduce_sum(tf.square(r))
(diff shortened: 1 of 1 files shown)


edward/inferences/hmc.py
```diff
@@ -76,9 +76,9 @@ def build_update(self):
     # Simulate Hamiltonian dynamics.
     new_sample = old_sample
     new_r_sample = old_r_sample
-    for _ in range(self.n_steps):
-      new_sample, new_r_sample = leapfrog(new_sample, new_r_sample,
-                                          self.step_size, self._log_joint)
+    new_sample, new_r_sample = leapfrog(new_sample, new_r_sample,
```
dustinvtran: cool! i think we can also remove the two lines above this now as well.

edward/inferences/hmc.py
```diff
@@ -152,18 +152,23 @@ def _log_joint(self, z_sample):
...
   grad_log_joint = tf.gradients(log_joint(z_old), list(six.itervalues(z_old)))
-  for i, key in enumerate(six.iterkeys(z_old)):
-    z, r = z_old[key], r_old[key]
-    r_new[key] = r + 0.5 * step_size * grad_log_joint[i]
-    z_new[key] = z + step_size * r_new[key]
-
-  grad_log_joint = tf.gradients(log_joint(z_new), list(six.itervalues(z_new)))
-  for i, key in enumerate(six.iterkeys(z_old)):
-    r_new[key] += 0.5 * step_size * grad_log_joint[i]
+  for n in range(n_steps):
```
dustinvtran: since we're not using the looping variable, we should do `for _ in ...` for readability

edward/inferences/hmc.py
```diff
@@ -152,18 +152,23 @@ def _log_joint(self, z_sample):
     return log_joint
 
 
-def leapfrog(z_old, r_old, step_size, log_joint):
+def leapfrog(z_old, r_old, step_size, log_joint, n_steps):
   z_new = {}
   r_new = {}
 
+  for key in z_old:
```
dustinvtran: i think these lines can be condensed with the the copy method in dict's, e.g., `z_new = z_old.copy()` and `r_new = r_old.copy()`.

dustinvtran:
thanks for the catch; this looks great. only minor comments above

dustinvtran:
hi @matthewdhoffman: i made the minor changes from my comments above. let me know if you approve/have comments, and i'll merge.

matthewdhoffman:
LGTM, thanks for cleaning it up. (And for showing me a couple of better idioms!)

Issue #412, opened 2017-01-18
"Dimensions must be equal" Error with HMC&SGLD


mathetake:
Hi. I am new to here and stuck with some problems with HMC&SGLD.

I've tried to  develop a 4-layer neural network with Edward's HMC&SGLD inference.

My code for SGLD is the following;

```
def four_layer_nn(x,W_1,W_2,W_3,b_1,b_2):
    h = tf.tanh(tf.matmul(x, W_1) + b_1)
    h = tf.tanh(tf.matmul(h, W_2) + b_2)
    h = tf.matmul(h, W_3) 
    return tf.reshape(h, [-1])

N = X_train.shape[0]  # data points
D = X_train.shape[1] # feature
M = 50000 #batch size

W_1 = Normal(mu=tf.zeros([D, 20]), sigma=tf.ones([D, 20])*100)
W_2 = Normal(mu=tf.zeros([20, 15]), sigma=tf.ones([20, 15])*100)
W_3 = Normal(mu=tf.zeros([15, 1]), sigma=tf.ones([15, 1])*100)
b_1 = Normal(mu=tf.zeros(20), sigma=tf.ones(20)*100)
b_2 = Normal(mu=tf.zeros(15), sigma=tf.ones(15)*100)

#Empirical Variable for SGLD
qW_1 = Empirical(params=tf.Variable(tf.random_normal([M,D,20])))
qW_2 = Empirical(params=tf.Variable(tf.random_normal([M,20,15])))
qW_3 = Empirical(params=tf.Variable(tf.random_normal([M,15,1])))
qb_1 = Empirical(params=tf.Variable(tf.random_normal([M,20])))
qb_2 = Empirical(params=tf.Variable(tf.random_normal([M,15])))

x_ph  = tf.placeholder(tf.float32,[M,D])
y = Bernoulli(logits=four_layer_nn(x_ph, W_1, W_2, W_3, b_1, b_2))
y_ph = tf.placeholder(tf.float32,[M])
inference = ed.SGLD({W_1: qW_1, b_1: qb_1, W_2: qW_2 , b_2: qb_2, W_3: qW_3 }, data={y: y_ph})
inference.initialize(n_iter=n_epoch*int(float(N)/M), n_print=100)
...(Batch training codes follow.)
```

When I run this code, the following error occurs:
```
Traceback (most recent call last):
  File "dnn_SGLD.py", line 85, in <module>
[...]

PR #415, opened 2017-01-19
[WIP] Added implementation of SGHMC and example.

dwadden:
I'm working with @nfoti and @yianma. We have some questions on a few small parts of the code. See comments in changed files.

### DIFF
--- a/edward/inferences/sghmc.py
+++ b/edward/inferences/sghmc.py
@@ -1,0 +1,147 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import six
+import tensorflow as tf
+
+from edward.inferences.monte_carlo import MonteCarlo
+from edward.models import Normal, RandomVariable, Empirical
+from edward.util import copy
+
+
+class SGHMC(MonteCarlo):
+  """Stochastic gradient Hamiltonian Monte Carlo (Chen et al., 2014).
+
+  Notes
+  -----
+  In conditional inference, we infer :math:`z` in :math:`p(z, \\beta
+  \mid x)` while fixing inference over :math:`\\beta` using another
+  distribution :math:`q(\\beta)`.
+  ``SGHMC`` substitutes the model's log marginal density
+
+  .. math::
+
+    \log p(x, z) = \log \mathbb{E}_{q(\\beta)} [ p(x, z, \\beta) ]
+                \\approx \log p(x, z, \\beta^*)
+
+  leveraging a single Monte Carlo sample, where :math:`\\beta^* \sim
+  q(\\beta)`. This is unbiased (and therefore asymptotically exact as a
+  pseudo-marginal method) if :math:`q(\\beta) = p(\\beta \mid x)`.
+  """
+  def __init__(self, *args, **kwargs):
+    """
+    Examples
+    --------
+    >>> z = Normal(mu=0.0, sigma=1.0)
+    >>> x = Normal(mu=tf.ones(10) * z, sigma=1.0)
+    >>>
+    >>> qz = Empirical(tf.Variable(tf.zeros([500])))
+    >>> data = {x: np.array([0.0] * 10, dtype=np.float32)}
+    >>> inference = ed.SGHMC({z: qz}, data)
+    """
+    super(SGHMC, self).__init__(*args, **kwargs)
+
+  def initialize(self, step_size=0.25, friction=0.1, *args, **kwargs):
+    """
+    Parameters
+    ----------
+    step_size : float, optional
+      Constant scale factor of learning rate.
+    friction : float, optional
+      Constant scale on the friction term in the Hamiltonian system.
+    """
+    self.step_size = step_size
+    self.friction = friction
+    self.v = {z: tf.Variable(tf.zeros(qz.params.get_shape()[1:]))
+              for z, qz in six.iteritems(self.latent_vars)}
+    return super(SGHMC, self).initialize(*args, **kwargs)
+
+  def build_update(self):
+    """
+    Simulate Hamiltonian dynamics with friction using a discretized
+    integrator. Its discretization error goes to zero as the learning rate
+    decreases.
+    Implements the update equations from (15) of Chen et al., 2014.
+    """
+    old_sample = {z: tf.gather(qz.params, tf.maximum(self.t - 1, 0))
+                  for z, qz in six.iteritems(self.latent_vars)}
+    old_v_sample = {z: v for z, v in six.iteritems(self.v)}
+
+    # Simulate Hamiltonian dynamics with friction.
+    friction = tf.constant(self.friction, dtype = tf.float32)
+    learning_rate = tf.constant(self.step_size * 0.01, dtype = tf.float32)
+    grad_log_joint = tf.gradients(self._log_joint(old_sample),
+                                  list(six.itervalues(old_sample)))
+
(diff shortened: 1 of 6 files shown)


edward/inferences/sghmc.py
```diff
@@ -0,0 +1,152 @@
...
+    return super(SGHMC, self).initialize(*args, **kwargs)
+
+  def build_update(self):
+    """
+    Simulate Hamiltonian dynamics with friction using a discretized
+    integrator. Its discretization error goes to zero as the learning rate
+    decreases.
+    Implements the update equations from (15) of Chen et al., 2014.
+    """
+    # TODO: Would it be more memory-efficient to use tf.control_dependencies,
```
dwadden: Does making aliases make copies of the tensors? If so, would it save memory to use `tf.control_dependencies` and mutate the objects directly?

edward/inferences/sghmc.py
```diff
@@ -0,0 +1,152 @@
...
+    # rather than repeatedly re-assign attributes stored on self.v?
+    old_sample = {z: tf.gather(qz.params, tf.maximum(self.t - 1, 0))
+                  for z, qz in six.iteritems(self.latent_vars)}
+    old_v_sample = {z: v for z, v in six.iteritems(self.v)}
+
+    # Simulate Hamiltonian dynamics with friction.
+    friction = tf.constant(self.friction, dtype = tf.float32)
+    # TODO: Allow option for exponentially decaying learning rate, or similar.
```
dwadden: For SGLD, we found that using a constant learning rate performed better than using exponential decay. Perhaps this should be an option passed into the inference method?

edward/inferences/sghmc.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    Utility function to calculate model's log joint density,
+    log p(x, z), for inputs z (and fixed data x).
+
+    Parameters
+    ----------
+    z_sample : dict
+      Latent variable keys to samples.
+    """
+    # TODO: This appears to be identical across HMC, SGHMC, SGLD. Should it be
```
dwadden: The `_log_joint` function looks like it's the same. Refactor?

examples/normal_sghmc.py
```diff
@@ -0,0 +1,80 @@
...
+import edward as ed
+import tensorflow as tf
+import numpy as np
+from matplotlib import pyplot as plt
+plt.style.use("ggplot")
+
+from edward.models import Empirical, MultivariateNormalFull
+
+# Helper functions
+# TODO: Where should these go?
```
dwadden: I added some helper functions to visualize the contours of a bivariate Gaussian (or more generally, any Edward distribution with a pdf method). Should this be refactored?

examples/normal_sghmc.py
```diff
@@ -0,0 +1,80 @@
...
+    X, Y = np.meshgrid(xs, ys)
+    T = tf.convert_to_tensor(np.c_[X.flatten(), Y.flatten()], dtype=tf.float32)
+    Z = sess.run(tf.exp(z.log_prob(T))).reshape((len(xs), len(ys)))
+    if ax is None:
+        fig, ax = plt.subplots()
+    cs = ax.contour(X, Y, Z)
+    if label:
+        plt.clabel(cs, inline=1, fontsize=10)
+
+def retrieve_samples(qz):
```
dwadden: It took a bit of work to figure out how to grab the actual MCMC samples. Is there a method to do this that we missed, or should we move this elsewhere to allow for more general use?

dustinvtran: if you just want the samples you can do `qz.params`

edward/inferences/sghmc.py
dustinvtran: sure, that makes sense. note one difficulty i had when implementing SGLD was that i could not leverage any of the existing tensorflow optimizers, and if more learning rates were an option, one would need to implement them all from scratch. if there were a way to use the tensorflow optimizers, it would be nice so long as the implementation doesn't ruin the readability and/or speed of the code.

edward/inferences/sghmc.py
dustinvtran: i'm not sure i understood the question but for example:
```python
>>> x = tf.constant(0)
>>> y = x
>>> x, y
(<tf.Tensor 'Const:0' shape=() dtype=int32>, <tf.Tensor 'Const:0' shape=() dtype=int32>)
```
y has a reference to the same tensor and doesn't recreate it from scratch, if that's what you're asking.

edward/inferences/sghmc.py
dustinvtran: yeah, that sounds useful to refactor (although i'm not sure where the best location is to put it)

examples/normal_sghmc.py
dustinvtran: that's awesome. maybe we can leave it here for now and think about refactoring it in another PR that focuses more on general visualizations for edward

edward/inferences/sghmc.py
dwadden: Got it re: tensorflow.

I'll make a new PR that adds an option to keep the learning rate constant, along with an example showing cases where this may be a bit easier to work with.

edward/inferences/sghmc.py
dwadden: OK. Will think about this and submit another PR.

examples/normal_sghmc.py
dwadden: Cool.

examples/normal_sghmc.py
dwadden: Nice, thanks

dwadden:
Did a bit of code cleanup - fixed indentation to 2 spaces in `normal_sghmc.py` and got rid of the TODO's since we've now addressed how to handle them.
Made a minor mistake and squashed those commits into the previous one so that the original changes we discussed don't show up. Now I know not to do that.
Let me know if there's anything else you'd like changed. Otherwise I'll start on the PR for adaptive step sizes.

edward/inferences/sghmc.py
dwadden: Yep, that was the question. Just wanted to make sure it wasn't recreating.

dwadden:
Cleaned things up a bit. Got rid of the README notes since we've now discussed them. Squashed into previous commit, which I now know is a bad thing to do because the changes we discussed are gone. Lesson learned.

Let me know if there are other things you'd like changed on this PR. Otherwise, I'll start thinking about the other thing we discussed: step size, code viz, and refactor of _log_joint.

dustinvtran:
looks good to me. can you add an example using data? e.g., linear/logistic regression is fine.

dwadden:
Will do.

dwadden:
Git style question: I'm want to update this branch with the latest changes from master, which contains relevant bug fixes for #412 . I'm also going to add my integer type fix. 

In general, this can be done by merging master into this branch, or rebasing the branch onto master. The problem with the former approach is it will make the history messy, while doing the latter makes it impossible to view changes that were made before the rebase (like what's happened above).

Which do you prefer?

dustinvtran:
i personally prefer to rebase to keep the commit history clean. that said, we're always squashing+merging pull requests, so feel free to do either.

dwadden:
Cool, rebase works for me.

dustinvtran:
i ran the examples in tensorflow 0.11.0 with python 2 and 3. i also double-checked upon merging that they work in tensorflow 1.0.0rc0. merging now.

Issue #407, opened 2017-01-17
Unable to replace Categorical RandomVariable using a data dict during inference


jhaberstro:
The `Categorical` class expects that the parameter to `log_prob` be an integer type. When replacing a `Categorical` RV during inference with integer data:

```
data = [1, 2, 3]
x = Categorical(logits=..., validate_args=True)
...
inference = ed.KLqp({...},
                    {x: tf.constant(data)})
inference.run()
```

the following assert fires:

```
  File "lda.py", line 60, in <module>
    inference.run()
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/edward/inferences/inference.py", line 211, in run
    self.initialize(*args, **kwargs)
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/edward/inferences/klqp.py", line 66, in initialize
    return super(KLqp, self).initialize(*args, **kwargs)
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/edward/inferences/variational_inference.py", line 70, in initialize
    self.loss, grads_and_vars = self.build_loss_and_gradients(var_list)
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/edward/inferences/klqp.py", line 117, in build_loss_and_gradients
    return build_score_loss_and_gradients(self, var_list)
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/edward/inferences/klqp.py", line 567, in build_score_loss_and_gradients
    x_log_prob = tf.reduce_sum(x_copy.log_prob(dict_swap[x]))
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/tensorflow/contrib/distributions/python/ops/distribution.py", line 437, in log_prob
    return self._log_prob(value)
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/tensorflow/contrib/distributions/python/ops/categorical.py", line 143, in _log_prob
    return -nn_ops.sparse_softmax_cross_entropy_with_logits(logits, k)
  File "/Users/jhaberstro/anaconda/envs/datascience/lib/python3.5/site-packages/tensorflow/python/ops/nn_ops.py", line 778, in sparse_softmax_cross_entropy_with_logits
    precise_logits, labels, name=name)
[...]

PR #421, opened 2017-01-25
Fix type error in initializing TensorFlow computation graph for inference.

dwadden:
@nfoti and I discovered a bug when doing inference in a neural net with a softmax final layer (e.g. for multiclass classification). This PR fixes it.

### DIFF
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -132,8 +132,15 @@ class Inference(object):
               raise TypeError("Observed variable bindings do not have same "
                               "shape.")
 
-            # If value is a np.ndarray, store it in the graph.
-            ph = tf.placeholder(tf.float32, value.shape)
+            # If value is a np.ndarray, store it in the graph. Assign its
+            # placeholder to an appropriate data type.
+            if np.issubdtype(value.dtype, np.float):
+              ph_type = tf.float32
+            elif np.issubdtype(value.dtype, np.int):
+              ph_type = tf.int32
+            else:
+              raise TypeError("Data value has an unsupported type.")
+            ph = tf.placeholder(ph_type, value.shape)
             var = tf.Variable(ph, trainable=False, collections=[])
             self.data[key] = var
             sess.run(var.initializer, {ph: value})
--- a/tests/test-inferences/test_integer_variables.py
+++ b/tests/test-inferences/test_integer_variables.py
@@ -1,0 +1,50 @@
+"""Test that integer variables are handled properly during initialization."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import edward as ed
+import numpy as np
+import tensorflow as tf
+
+from edward.models import Normal, Categorical
+
+
+def neural_network(x, W_0, W_1, b_0, b_1):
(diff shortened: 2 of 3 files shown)


edward/inferences/inference.py
```diff
@@ -132,8 +132,15 @@ def __init__(self, latent_vars=None, data=None, model_wrapper=None):
               raise TypeError("Observed variable bindings do not have same "
                               "shape.")
 
-            # If value is a np.ndarray, store it in the graph.
-            ph = tf.placeholder(tf.float32, value.shape)
```
dwadden: The TensorFlow placeholder needs to have the same basic type (e.g. integer or float) as the numpy array for which it's serving as a placeholder.

tests/test-inferences/test_integer_variables.py
```diff
@@ -0,0 +1,50 @@
...
+                  sigma=softplus(tf.Variable(tf.random_normal([D, n_hidden]))))
+    qW_1 = Normal(mu=tf.Variable(tf.random_normal([n_hidden, K])),
+                  sigma=softplus(tf.Variable(tf.random_normal([n_hidden, K]))))
+    qb_0 = Normal(mu=tf.Variable(tf.random_normal([n_hidden])),
+                  sigma=softplus(tf.Variable(tf.random_normal([n_hidden]))))
+    qb_1 = Normal(mu=tf.Variable(tf.random_normal([K])),
+                  sigma=softplus(tf.Variable(tf.random_normal([K]))))
+    inference = ed.KLqp({W_0: qW_0, b_0: qb_0, W_1: qW_1, b_1: qb_1},
+                        data={y: y_train})
+    inference.run()
```
dwadden: Without the patch in this PR, this line fails with error:
`typeError: DataType float32 for attr 'Tlabels' not in list of allowed values: int32, int64`

dustinvtran:
perfect. this also fixes #407. i ran it locally and it looks great. (i think you were right to not also do the casting at line 156 when `isinstance(key, str)`. this is the case for model wrappers, and in their implementation we force everything to be `float32` by convention.)

not sure how to interpret that build error.

dwadden:
Huh yeah it works locally for me as well... weird build error. I'll take a look.

dustinvtran:
man, tensorflow is weird. the unit test succeeds if it's run on the file separately, but fails if it also runs `test_bayesian_nn.py`. the failure is always whichever one is run second. i'm not sure what sort of artifact is remaining in one graph vs the other.

dwadden:
Haha well that's not good. Do you think there's something I can do on my end to fix it?

dustinvtran:
i spotted the fix. it looks like if you add the line `sess = tf.InteractiveSession()` to the top of `test_integer_variables.py`, the test passes. the error arises when the session in `test_integer_variables.py` tries to use the session built in `test_bayesian_nn.py`; resetting the default session by calling `tf.InteractiveSession()` breaks this interaction.

i recommending adding `sess = tf.InteractiveSession()` to the top of the `test_integer(self)` method (i.e., line 23), and for consistency to do the same for `test_bayesian_nn.py`.

dwadden:
Nice. OK, will do.

dwadden:
OK, looks like things work now.

PR #374, opened 2016-12-10
Make edward compatible with Tensorflow 0.12

gokceneraslan:
Initial attempt to add TF 0.12 compatibility. 

I have changed a few things in order to make edward compatible with TF 0.12:

- `tf.initialize_all_variables()` calls have been renamed to `tf.global_variables_initializer()`

- `tf.Variable.ref()` calls are removed. See: https://github.com/tensorflow/tensorflow/commit/c46f70c06985edce4905668a943801040c7a1354#diff-ae1a8f7b66539f000615a4ab7e4b2151R460. `ref()` is an internal function now.

- `tf.GraphKeys.VARIABLES` is replaced with `tf.GraphKeys.GLOBAL_VARIABLES`.

As a side note, even after applying these changes, edward still doesn't work because of the bug introduced in TF 0.12rc0, see: https://github.com/tensorflow/tensorflow/issues/5930 and https://github.com/tensorflow/tensorflow/commit/77cfa97ac784212130e97bc6fa9e6f8acfc86c08 for the fix. Therefore I had to use TF master branch.

There are still 2 test failing, I'm looking into them.

Issue #367, opened 2016-11-30
Variable.ref() no longer exists as of TensorFlow v0.12


shoyer:
It's now a [private method `Variable._ref()`](https://github.com/tensorflow/tensorflow/blob/r0.12/tensorflow/python/ops/variables.py#L460))

Edward uses this in `edward/inferences/klpq.py ` and `edward/inferences/klqp.py` for computing gradients, in lines that look like:
```python
    grads = tf.gradients(
        -tf.reduce_mean(q_log_prob * tf.stop_gradient(w_norm)),
        [v.ref() for v in var_list])
```

I believe it should work if `[v.ref() for v in var_list]` is simply replaced by `var_list`.

dustinvtran:
the proposed solution makes sense. we should check this; i forget why but i recall having to add `ref()` for a reason. maybe it was because `tf.gradients()` didn't work explicitly on tensorflow variables and do such a conversion internally? (vague recollection)

gokceneraslan:
I addressed this and a few other TF 0.12 incompatibilities in #374.

dustinvtran:
looking back, i used `[v.ref() for v in var_list]` instead of `var_list` because the `compute_gradients()` method in TensorFlow optimizers does it: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/training/optimizer.py#L347

this looks like overkill for us so indeed we can drop it.

PR #426, opened 2017-01-27
[WIP] Upgrade to TensorFlow 1.0.0

### LINKED ISSUE #425: edward should upgrade to work with tensorflow 1.0.0rc0
currently it is not. __please continue to use 0.11.0 for now.__

this is a high priority to fix.

dustinvtran:
This pull request upgrades Edward's TensorFlow syntax to be compatible with TensorFlow 1.0.0-alpha and above. It is a work in progress.

This builds off #374.

fixes #425.


dustinvtran:
I ran all scripts in `examples/` using TensorFlow 1.0.0rc0. They all work excluding `tf_convolutional_vae.py` and `vae_convolutional_prettytensor.py`. These two use PrettyTensor, which hasn't upgraded to work with TensorFlow 1.0.0 yet.

One note (#367) is that we do require variable `ref()` calls in files such as `klqp.py`. For example, running `getting_started_example.py` without the `ref()`'s would not learn anything. I'm not sure why this is the case; perhaps this can be diagnosed in a future PR. For now, I changed it all to use the private method `_ref()`.

PR #428, opened 2017-01-30
Require TensorFlow 1.0.0rc0 instead of 1.0.0a0

rilut:
**EDIT:**  
~~While installing Edward via setup.py/pip on Windows is fine. On Linux, pip complains that TensorFlow 1.0.0rc0 isn't compatible with 1.0.0a0~~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ setup(
     author_email="dustin@cs.columbia.edu",
     packages=['edward', 'edward.criticisms', 'edward.inferences',
               'edward.models', 'edward.stats', 'edward.util'],
-    install_requires=['tensorflow>=1.0.0a0',
+    install_requires=['tensorflow>=1.0.0rc0',
                       'numpy>=1.7',
                       'six>=1.10.0'],
     extras_require={'stan': ['pystan>=2.0.1.3'],


mariru:
@rilut I Installed tensorflow-1.0.0rc0 on linux using pip with the instructions here
https://github.com/tensorflow/tensorflow/blob/r1.0/tensorflow/g3doc/get_started/os_setup.md
after that I was able to pip install edward without problems. (Note: I am using python 2.7, no GPUs)

pwl:
I have the same issue. When running `pip install edward==1.2.1` I'm getting
```
Collecting edward==1.2.1
  Downloading edward-1.2.1.tar.gz
Collecting tensorflow>=1.0.0a0 (from edward==1.2.1)
  Could not find a version that satisfies the requirement tensorflow>=1.0.0a0 (from edward==1.2.1) (from versions: 0.12.0rc0, 0.12.0rc1, 0.12.0, 0.12.1)
No matching distribution found for tensorflow>=1.0.0a0 (from edward==1.2.1)
```
This is happens after installing TF `1.0.0rc0` with `pip install https://storage.googleapis.com/tensorflow/linux/gpu/tensorflow_gpu-1.0.0rc0-cp35-cp35m-linux_x86_64.whl`.

rilut:
I'm starting to think: or is it because pip doesn't recognize `tensorflow-gpu` as `tensorflow`?

Is it possible to make pip/setup.py recognize both `tensorflow` and `tensorflow-gpu`?

pwl:
@rilut yep, that seems to be the case, I had no trouble installing it with `https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.0.0rc0-cp35-cp35m-linux_x86_64.whl`, so the non-gpu version of TF.

dustinvtran:
that's strange.. from looking at other packages, prettytensor seems to make a [similar remark here](https://github.com/google/prettytensor/blob/master/setup.py#L21). i'm also not sure if `install_requires` accepts booleans like `tensorflow>=1.0.0a0 or `tensorflow-gpu>=1.0.0a0`.

gokceneraslan:
I think there is just no way to handle it properly. One idea is to detect whether there is a GPU during installation in setup.py:

```python
from subprocess import check_output, CalledProcessError

try:
    num_gpus = len(check_output(['nvidia-smi', '--query-gpu=gpu_name',
                                 '--format=csv']).decode().strip().split('\n'))
    tf = 'tensorflow-gpu' if num_gpus > 1 else 'tensorflow'
except CalledProcessError:
    tf = 'tensorflow'

setup(..., install_requires=[tf, ...])
```

But it's just an ugly hack.

dustinvtran:
In the mean time I submitted an issue on TensorFlow (https://github.com/tensorflow/tensorflow/issues/7166).

PR #433, opened 2017-02-01
FIX #430: added pillow to setup.py dependencies

### LINKED ISSUE #430: Add PIL package to dependencies
The `PIL` package is not installed by default with `scipy` but it is necessary to run the `vae.py` example. It should probably be added to the visualization dependencies in `setup.py`.

nfoti:
Fixes #430


dustinvtran:
lgtm

PR #434, opened 2017-02-01
create img/ dir in vae example

nfoti:
The script `examples/vae.py` assumes that there is a directory `img/`, but this is kind of annoying, so this PR makes the directory automatically.

### DIFF
--- a/examples/vae.py
+++ b/examples/vae.py
@@ -1,13 +1,12 @@
 #!/usr/bin/env python
 """Variational auto-encoder for MNIST data.
-
-Assumes the directories "img/" and "data/mnist/" exist.
 """
 from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
 import edward as ed
+import os.path
 import tensorflow as tf
 
 from edward.models import Bernoulli, Normal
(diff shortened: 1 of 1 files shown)


examples/vae.py
```diff
@@ -17,6 +19,11 @@
 from scipy.misc import imsave
 from tensorflow.examples.tutorials.mnist import input_data
 
+try:
```
dustinvtran: instead of try/except, I find [this approach](https://github.com/blei-lab/edward/blob/master/examples/tf_convolutional_vae.py#L101-L107) a bit more readable, and also letting `IMG_DIR` be a variable.

nfoti:
I updated the code to use the style that @dustinvtran suggested. I checked if there were any other examples that could use this fix but this seems to have been the only example that didn't create the `img/` directory if it was missing.

dustinvtran:
cool! can you remove the blankline in the import statement (sorted in alphabetical order, so edward -> os -> tensorflow) and delete the assumption line?

nfoti:
I fixed the import lines you mentioned. I don't know that the 'assumption line' you refer to is though. I'm happy to fix it once I know though.

dustinvtran:
line 4; the making of the data directory makes it not required

nfoti:
done.

dustinvtran:
thanks again!

PR #437, opened 2017-02-03
Fixed broken LaTeX equation (missing defs)

jamestwebber:


dustinvtran:
good catch. this error came up as text was synchronized between the documentation and the white paper.

PR #440, opened 2017-02-05
[DOC]: another latex hiccup

jamestwebber:
Since I'm still reading the docs, figured I might as well.


dustinvtran:
yes, these are all welcome. thanks again!

jamestwebber:
I'm not familiar with how the templating works but I _think_ that there was some bad indentation toward the bottom of the file, so I just fixed that as well.

PR #456, opened 2017-02-14
Generalize KL penalties in KLqp to go beyond the normal distribution

dustinvtran:
Title. We use the KL method available in `tf.contrib.distributions`, which is registered for a number of distributions. This avoids duplicate effort from having to write our own functions. Using this KL function also lets us apply other analytic KL penalties during variational inference.

Remarks
+ In generalizing the KL penalties, I removed this `inference.data.get(z, 1.0)` term, which I'm not sure why it exists. It don't think it would ever be `z` instead of 1.0. I need to investigate this further.

todo
+ [ ] In `KLqp`, we need to determine whether the KL is registered for a pair of distributions. How?


edward/inferences/klqp.py
```diff
@@ -92,25 +93,25 @@ def build_loss_and_gradients(self, var_list):
...
     is_reparameterizable = all([rv.is_reparameterized and rv.is_continuous
                                 for rv in six.itervalues(self.latent_vars)])
-    qz_is_normal = all([isinstance(rv, Normal) for
-                       rv in six.itervalues(self.latent_vars)])
-    z_is_normal = all([isinstance(rv, Normal) for
-                       rv in six.iterkeys(self.latent_vars)])
-    is_analytic_kl = qz_is_normal and \
-        (z_is_normal or hasattr(self.model_wrapper, 'log_lik'))
+    # TODO(dt): determine if there is a kl registered for kl(qz, z)
+    # for each pair in self.latent_vars. how?
```
shoyer: Looks like the best you can do for now is catch `NotImplementedError` when calling `distributions.kl`. I guess you could make a helper function to do the function call, but it's pretty awkward and will help some junk in the graph. This is probably something worth adding to TensorFlow. CC @ebrevdo

ebrevdo:
Do you need to know if the KL is registered only at the point when you want to calculate it?  If so, just catching a NotImplementedError is enough.

dustinvtran:
Unfortunately no. In the above, there are a list of `build_loss_and_gradients` functions, one of which uses analytic KL penalties. We'd like to detect KL and then apply that `build_loss_and_gradients` function. (A more local version of `build_loss_and_gradients` is being pursued, separate from this though.)

I think @shoyer's proposal is an okay temporary solution. As he notes though, it adds a lot of the junk to the graph.

PR #460, opened 2017-02-15
Scaled kl

vrama91:
Implements scaling of KL terms in ELBO (when computed using the KL form) by some user provided constants.

### DIFF
--- a/edward/inferences/klqp.py
+++ b/edward/inferences/klqp.py
@@ -426,9 +426,8 @@ def build_reparam_kl_loss_and_gradients(inference, var_list):
   p_log_lik = tf.stack(p_log_lik)
 
   if inference.model_wrapper is None:
-    kl = tf.reduce_sum([inference.data.get(z, 1.0) *
-                        tf.reduce_sum(kl_multivariate_normal(
-                            qz.mu, qz.sigma, z.mu, z.sigma))
+    kl = tf.reduce_sum([inference.kl_scaling.get(z, 1.0) *
+                        tf.reduce_sum(ds.kl(qz, z))
                         for z, qz in six.iteritems(inference.latent_vars)])
   else:
     kl = tf.reduce_sum([tf.reduce_sum(kl_multivariate_normal(qz.mu, qz.sigma))
@@ -643,9 +642,9 @@ def build_score_kl_loss_and_gradients(inference, var_list):
   q_log_prob = tf.stack(q_log_prob)
 
   if inference.model_wrapper is None:
-    kl = tf.reduce_sum([inference.data.get(z, 1.0) *
-                        tf.reduce_sum(kl_multivariate_normal(
-                            qz.mu, qz.sigma, z.mu, z.sigma))
+
+    kl = tf.reduce_sum([inference.kl_scaling.get(z, 1.0) *
+                        tf.reduce_sum(ds.kl(qz, z))
                         for z, qz in six.iteritems(inference.latent_vars)])
   else:
     kl = tf.reduce_sum([tf.reduce_sum(kl_multivariate_normal(qz.mu, qz.sigma))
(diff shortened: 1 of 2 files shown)


dustinvtran:
This is great work—thanks Rama. The approach with the dictionary makes sense to me.

For the `kl_scaling` argument, is there a reason you prefer it in `VariationalInference` over `KLqp`? KL penalties are easy to do for `KLqp` since (up to constants) it can be written as the sum of an expected log-likelihood and a bunch of KLs between q and the prior. I'm not sure if that's applicable for other objectives.

Also, what do you think about putting it in `initialize()` instead of `__init__()`?

vrama91:
Thanks for the comments Dustin. About `kl_scaling`, I thought of putting it in `VariationalInference` since a lot of other classes in `klqp.py` such as `ReparameterizationKLKLqp` and `ScoreKLKLqp` can also technically be used with `kl_scaling`. These classes all inherit from `VariationalInference` so my choice of where to put `kl_scaling` was more or less a practical one of not having to change the code for all these related classes independently.
However, if you think I should move it I can do that. As you say, `kl_scaling` might not be applicable for other VI objectives. Let me know.

Sure, I can move `kl_scaling` to `initialize()` instead of `__init__()`.

dustinvtran:
That's a good point. I guess it's similar to where to put the `n_samples` argument.

For now I think it makes sense to duplicate that argument in `initialize()` for all the `KLqp` versions that would apply it: `KLqp`, `ReparameterizationKLKLqp`, `ScoreKLKLqp`. We can think later about more abstractions when the redundancy is too much.

vrama91:
Makes a lot of sense. I made the changes. Let me know if you have any other thoughts / feedback!

edward/inferences/klqp.py
```diff
@@ -643,9 +669,9 @@ def build_score_kl_loss_and_gradients(inference, var_list):
   q_log_prob = tf.stack(q_log_prob)
 
   if inference.model_wrapper is None:
-    kl = tf.reduce_sum([inference.data.get(z, 1.0) *
-                        tf.reduce_sum(kl_multivariate_normal(
-                            qz.mu, qz.sigma, z.mu, z.sigma))
+
```
dustinvtran: emptyline, which can be removed

edward/inferences/variational_inference.py
```diff
@@ -19,7 +19,7 @@
 class VariationalInference(Inference):
   """Base class for variational inference methods.
   """
-  def __init__(self, *args, **kwargs):
+  def __init__(self,  *args, **kwargs):
```
dustinvtran: pep8 might break due to this extra space

edward/inferences/klqp.py
```diff
@@ -53,16 +53,25 @@ class KLqp(VariationalInference):
   def __init__(self, *args, **kwargs):
     super(KLqp, self).__init__(*args, **kwargs)
 
-  def initialize(self, n_samples=1, *args, **kwargs):
+  def initialize(self, n_samples=1, kl_scaling={}, *args, **kwargs):
```
dustinvtran: instead of the default argument being an empty dict, i recommend using `None`. also see http://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

this is a bit safer. and in the function, you can write
```python
if kl_scaling is None:
  kl_scaling = {}
```

vrama91: Thanks! This is a really useful tip.

dustinvtran:
in the rebase i think you forgot to revert the `ds.kl` calls back to the `kl_multivariate_normal` stuff. this would also fix the travis build failure.

vrama91:
Thanks for spotting this! Corrected it, but this doesn't seem to have fixed the travis build issue. Am I missing something?

dustinvtran:
you should swap the args for `kl_multivariate_normal`.
```diff
-                        tf.reduce_sum(kl_multivariate_normal(qz, z))
+                        tf.reduce_sum(kl_multivariate_normal(
+                            qz.mu, qz.sigma, z.mu, z.sigma))
```

vrama91:
Sorry about that! Thanks for iterating with me on this :-)

Issue #453, opened 2017-02-12
Mixed Effects Model GLMM Examples


patrickeganfoley:
Hello!  

Would there be any interest in adding some canonical mixed effects model examples?  Maybe something using classic datasets like InstEval (student ratings of instructors, data in the R lme4 package) or the Radon dataset (like @twiecki 's [pym3](http://twiecki.github.io/blog/2014/03/17/bayesian-glms-3/) example).  

They are all pretty simple, but I think having something showing how to use `tf.gather` for id lookups would be helpful.

dustinvtran:
sure, that would be a welcome contribution. none of the current bayesian regression examples are hierarchical (unless you count the [rasch model](https://github.com/blei-lab/edward/blob/master/examples/rasch_model_hmc.py))), so it would also be helpful for that reason. 

with standard normal random effects, it would be very easy to do fast and scalable variational inference. HMC would also work.

patrickeganfoley:
Cool!  Should I keep datasets self contained and simulated, or can I add a csv to `/examples/data` ?

dustinvtran:
real data is preferred. feel free to add it to that directory if there's not a function in another common package that can load it.

patrickeganfoley:
Just opened https://github.com/blei-lab/edward/pull/461 to check if I am in the right direction.  Tried to include some form of criticism (visualization!) for both.

PR #461, opened 2017-02-15
WIP - Mixed Effects Models Examples

patrickeganfoley:
Hello!   Going off of https://github.com/blei-lab/edward/issues/453 , I made some simple examples of mixed effects models.  

There are two models:
  * irt.py - implements a classic Item Rating Theory model with simulated data
  * linear_mixed_effects_model.py - Fits a standard example model for predicting students' ratings of professors using the `InstEval` data set from the R package `lme4`, which I have included in `/examples/data/`.  

Is this basically in the right direction?


dustinvtran:
> Is this basically in the right direction?

Yep, this looks great. I like that you include the lmer formula too. A few pieces of advice:

1. We use PEP 8 style with two-space indents, which you can check by running, e.g., `pep8 irt.py` in terminal (pep8 is a [pip installable package](https://pypi.python.org/pypi/pep8)).

2. I noticed you're building ops for the computation graph in the for loop. You can get much faster speedup by building the ops outside the loop, and then calling them. For example, instead of writing `student_etas_post = qstudents.mean().eval()` in the loop, you can write `qstudents_mean = qstudents.mean()` and then call `student_etas_post = qstudents_mean.eval()` in the loop.

3. Is there a typical data set one can use for the IRT model?

patrickeganfoley:
@dustinvtran 

Great!  Thanks for the help!

1.  Done, now with two spaces.  👍 

2.  Thanks!  Switched those out and it does seem faster!  

3.  I've looked around a bit, and I haven't yet found any that are clearly open source.  There is `masc1` referenced [here by a Stata blog](http://blog.stata.com/2016/01/18/bayesian-binary-item-response-theory-models-using-bayesmh/) but I don't think it's free.  I can look around a bit more, there are some others that _might_ be open, but I haven't yet found documentation.

examples/linear_mixed_effects_model.py
```diff
@@ -0,0 +1,160 @@
...
+qdept_mean = q_eta_dept.mean()
+qmu_mean = qmu.mean()
+qservice_mean = qservice.mean()
+
+for t in range(inference.n_iter):
+  info_dict = inference.update()
+  inference.print_progress(info_dict)
+
+  if t % inference.n_print == 0:
+      y_post = ed.copy(y, {mu: qmu_mean,
```
dustinvtran: you can move this line outside the loop since it only needs to be written once.

examples/linear_mixed_effects_model.py
```diff
@@ -0,0 +1,160 @@
...
+                       eta_d: qd_mean,
+                       eta_dept: qdept_mean})
+
+      eta_s_post = qs_mean.eval()
+      eta_d_post = qd_mean.eval()
+      eta_dept_post = qdept_mean.eval()
+      mu_post = qmu_mean.eval()
+      service_post = qservice_mean.eval()
+
+      service_X_test = tf.placeholder(tf.float32, [n_obs_test, 1])
```
dustinvtran: same as above; you can move this line, along with `yhat_test` below, outside the loop. the generla idea is that you don't want to be adding nodes to the graph at runtime; with a static graph, computation is super fast.

dustinvtran:
Nice work. Only minor comments above. If there's no real data set available for the IRT model that's okay.

patrickeganfoley:
Cool!  Just to make sure I understand - I want the definitions outside, but it is OK to have `x.eval()` statements within loops for printing or displaying things?

dustinvtran:
Yep. The declarations such as `ed.copy()` and `tf.gather` can be done once, as they build nodes in the graph. The loop can then simply evaluate stuff by calling `sess.run()` and `eval()` on these nodes.

patrickeganfoley:
Thanks!  Just made those changes.

dustinvtran:
I ran the scripts. All looks great—love the visual animations during training. Thanks again for this.

patrickeganfoley:
Great!  Thanks again for all your help!  

Just closed the issue https://github.com/blei-lab/edward/issues/453 .

PR #462, opened 2017-02-15
A due reshape of y_train and y_test to avoid errors in the inference …

LukeMathWalker:
Without reshaping y_train and y_test the following error arises:
```
File "MixtureDensityNetwork.py", line 120, in <module>
    info_dict = inference.update(feed_dict={X_ph: X_train, y_ph: y_train})
[...]
ValueError: Cannot feed value of shape (4000, 1) for Tensor 'Placeholder_1:0', which has shape '(?,)'
```

### DIFF
--- a/examples/mixture_density_network.py
+++ b/examples/mixture_density_network.py
@@ -82,6 +82,8 @@ K = 20  # number of mixture components
 
 # DATA
 X_train, X_test, y_train, y_test = build_toy_dataset(N)
+y_train = y_train.reshape(-1)
+y_test = y_test.reshape(-1)
 print("Size of features in training data: {}".format(X_train.shape))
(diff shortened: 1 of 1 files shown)


dustinvtran:
Interesting. I checked Python 2 and 3, and this works fine without the explicit reshaping. Do you know what line makes `y_{train,test}` a matrix?

LukeMathWalker:
It turns out I was using a different (the old?) version of the function `build_toy_dataset` where the following line turns `y_{train,test}` into a matrix:
```
y_data = np.random.uniform(-10.5, 10.5, (N, 1)).astype(np.float32)
```

dustinvtran:
ah, got it. thanks for verifying anyways!

PR #466, opened 2017-02-17
dirichlet_categorical.py

bertini36:
Dirichlet-Categorical model


dustinvtran:
nice work. this looks useful. to keep it consistent with other example scripts, can you replace those lines with
```python
#!/usr/bin/env python
```
and then also include
```
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```

bertini36:
Done!

dustinvtran:
thanks

PR #468, opened 2017-02-20
fix missing case to feed np.float variables

nfoti:
Creates a new path in `inference.py` to handle the case where we feed in `np.float` variables to nodes in the computation graph.

### DIFF
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -151,6 +151,30 @@ class Inference(object):
                               "shape.")
 
             self.data[key] = value
+          elif isinstance(value, np.number):
+            if np.issubdtype(value.dtype, np.float):
+              ph_type = tf.float32
+            elif np.issubdtype(value.dtype, np.int):
+              ph_type = tf.int32
+            else:
+                raise TypeError("Data value as an invalid type.")
+            ph = tf.placeholder(ph_type, value.shape)
+            var = tf.Variable(ph, trainable=False, collections=[])
+            self.data[key] = var
+            sess.run(var.initializer, {ph: value})
+          elif isinstance(value, float):
+            ph_type = tf.float32
+            ph = tf.placeholder(ph_type, ())
+            var = tf.Variable(ph, trainable=False, collections=[])
+            self.data[key] = var
+            sess.run(var.initializer, {ph: value})
+          elif isinstance(value, int):
+            ph_type = tf.int32
+            ph = tf.placeholder(ph_type, ())
+            var = tf.Variable(ph, trainable=False, collections=[])
+            self.data[key] = var
+            # handle if value is `bool` which this case catches
+            sess.run(var.initializer, {ph: int(value)})
           else:
             raise TypeError("Data value has an invalid type.")
         elif isinstance(key, str):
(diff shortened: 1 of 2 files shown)


edward/inferences/inference.py
```diff
@@ -151,6 +151,13 @@ def __init__(self, latent_vars=None, data=None, model_wrapper=None):
                               "shape.")
 
             self.data[key] = value
+          elif isinstance(value, np.float64):
```
dustinvtran: cool. is there a way to get this work for non-float64 types too? similar to [`L137-140`](https://github.com/nfoti/edward/blob/2970e4b71270225dcd5596d38e4bab646470b38d/edward/inferences/inference.py#L137-L140), so that you can enter into this clause if it's int or float.

dustinvtran:
only one comment above. can you also add a line to the `test_data` function in `tests/test-inferences/test_inference.py`? you can check for example that instantiating `ed.Inference` works when the data passed in as a float.

nfoti:
this version handles both numpy floats and ints (32- and 64-bit) and added the requested test.

nfoti:
this newest version also handles python scalars (including `bool`s).

nfoti:
hmm, i'm not sure why the test fails. it passes on my laptop.

dustinvtran:
The same issue appeared [here](https://github.com/blei-lab/edward/pull/421#issuecomment-275027384). I think this is an issue with `test_bayesian_nn.py` interacting with other tests. Can you try rebasing/merging to the most recent commit? I think that'll fix it.

dustinvtran:
oh, actually the below fixes it
```diff
--- a/tests/test-inferences/test_inference.py
+++ b/tests/test-inferences/test_inference.py
@@ -20,6 +20,7 @@ class NormalNormal:
 class test_inference_class(tf.test.TestCase):

   def test_latent_vars(self):
+    tf.InteractiveSession()
     mu = Normal(mu=0.0, sigma=1.0)
     qmu = Normal(mu=tf.Variable(0.0), sigma=tf.constant(1.0))
     qmu_misshape = Normal(mu=tf.constant([0.0]), sigma=tf.constant([1.0]))
@@ -31,6 +32,7 @@ class test_inference_class(tf.test.TestCase):
     self.assertRaises(TypeError, ed.Inference, {mu: qmu_misshape})

   def test_data(self):
+    tf.InteractiveSession()
     x = Normal(mu=0.0, sigma=1.0)
     qx = Normal(mu=0.0, sigma=1.0)
     qx_misshape = Normal(mu=tf.constant([0.0]), sigma=tf.constant([1.0]))
@@ -47,6 +49,7 @@ class test_inference_class(tf.test.TestCase):
     self.assertRaises(TypeError, ed.Inference, data={x: qx_misshape})

   def test_model_wrapper(self):
+    tf.InteractiveSession()
     model = NormalNormal()
     qmu = Normal(mu=tf.Variable(0.0), sigma=tf.constant(1.0))
```
which you can check by running `py.test tests/test-inferences`.

nfoti:
should be fixed now.

dustinvtran:
thanks!

PR #469, opened 2017-02-20
Dirichlet-Categorical example is python3 compatible

rw:
Change `xrange` to `range` for python3 compatibility.


dustinvtran:
nice catch

PR #475, opened 2017-02-22
Invgamma-Normal Metropolis Hastings

bertini36:


dustinvtran:
Cool! Funny that the acceptance rate is so low. That proposal is close to the posterior though. Is Inverse Gamma just that pathological?

Looks like the Travis build failed due to a [PEP8 error](https://travis-ci.org/blei-lab/edward/jobs/204315588#L1200). In keeping convention with the other example scripts, I also recommend ending decimals with 0s, e.g., `7.0` instead of `7.` and removing significant digits, e.g., `0.7` instead of `0.70`. As a tip, you can also use `tf.zeros(1000)` which is slightly cleaner than `tf.zeros([1000])`.

dustinvtran:
from travis: https://travis-ci.org/blei-lab/edward/jobs/204383655#L1197

dustinvtran:
thanks again.

Issue #306, opened 2016-10-18
``edward.util.copy`` does not support tensorflow graphs with scan


bayerj:
Consider the following minimal example:

```
import numpy as np
import tensorflow as tf
from edward.util import copy

x = tf.constant(np.arange(10, dtype='float32'))
summed = tf.scan(lambda prev, cur: prev + cur, x)
copy(summed)
```

It raises an exception as follows:

```
*snip*
  File "/home/bayerj/devel/edward/edward/util/random_variables.py", line 175, in copy
    new_op = copy(op, dict_swap, scope, True, copy_q)
  File "/home/bayerj/devel/edward/edward/util/random_variables.py", line 207, in copy
    elem = copy(x, dict_swap, scope, True, copy_q)
  File "/home/bayerj/devel/edward/edward/util/random_variables.py", line 175, in copy
    new_op = copy(op, dict_swap, scope, True, copy_q)
  File "/home/bayerj/devel/edward/edward/util/random_variables.py", line 206, in copy
    for x in op.inputs:
  File "/home/bayerj/anaconda2/lib/python2.7/site-packages/tensorflow/python/framework/ops.py", line 1512, in inputs
    return Operation._InputList(self)
RuntimeError: maximum recursion depth exceeded
```

The internals of both `copy` and tensorflow are currently to obscure to me to look into this myself. 

To me it felt best as if such a function was part of tensorflow itself to keep up with new features. There is also an issue for that: https://github.com/tensorflow/tensorflow/issues/1070, but it has rather low activity.

dustinvtran:
fixed with #500

PR #480, opened 2017-02-27
WIP: examples of simple time series models

nfoti:
I have written a few examples of using edward to fit some simple time series models. They are in Jupyter notebook form right now, but I'm happy to convert them to scripts if that's easier (though I think having them as notebooks with some annotations can also be useful for learning edward). Any suggestions for improvement are welcome. I am planning on adding at least two more examples, one that uses a VAR(1) model for multiple replicates of a series and another example demonstrating a VAR(p) process (p > 1).

One point to notice is that constructing the `Inference` objects takes a very long time. I haven't dug into why this is happening, but I'm guessing that there is a lot of copying going when when I connect the training data to the corresponding nodes in the computation graph. Now that ICML is over I should have some time to dig into what's actually happening during construction and ways that we could potentially speed it up.

### DIFF
--- a/examples/time-series/example_ar-p.ipynb
+++ b/examples/time-series/example_ar-p.ipynb
@@ -1,0 +1,285 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2017-02-24T10:29:35.964814",
+     "start_time": "2017-02-24T10:29:35.600359"
+    },
+    "collapsed": false,
+    "inputHidden": false,
+    "outputHidden": false
+   },
+   "outputs": [],
+   "source": [
+    "from __future__ import absolute_import\n",
+    "from __future__ import division\n",
+    "from __future__ import print_function\n",
+    "\n",
+    "import edward as ed\n",
+    "import numpy as np\n",
+    "import tensorflow as tf\n",
+    "\n",
+    "import matplotlib.pyplot as plt\n",
+    "%matplotlib inline"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Generate data from stable AR(2) process\n",
+    "---"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "ExecuteTime": {
+     "end_time": "2017-02-24T10:29:36.820467",
+     "start_time": "2017-02-24T10:29:36.814343"
+    },
+    "collapsed": false,
+    "inputHidden": false,
+    "outputHidden": false
+   },
+   "outputs": [],
+   "source": [
+    "# Parameters\n",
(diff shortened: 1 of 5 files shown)


dustinvtran:
Cool! Good idea on the notebooks. I think this is the more suitable platform longer term as well. I'm looking into porting all tutorials and selected examples into notebooks.

re:`Inference`. Graph construction time scales roughly as the number of nodes in the model's computational graph grows. You could probably vectorize operations such as
```python
beta = [Normal(mu=0., sigma=2.) for i in xrange(p)]
```
to be a vector of normals, and then apply `tf.gather(beta, j)` to index it. (Of course, for only few parameters, it's the time series that's the bigger problem and not the beta.) We should definitely look into optimizing the graph construction; I think the copying done under the hood can be done faster and fewer times.

nfoti:
Do you think it's worth trying to specify the models with `tf.while_loop`? I tried to do this first but ran into some trouble, but not that I know more about `tf` and edward I may be successful.

dustinvtran:
That would definitel speed up construction time although at the cost of a big drag at run time. ( its also a bit harder to specify variatiaonal approximations of random varuables inside body functions)

Since you're using reparameterization gradients and/or MAP, the graph should only need to copy the likelihood in order to swap dependence to the approximate posterior instead of the prior. You can also try diagnosing the graph from TensorBoard by specifying the logdir argument in inference.

nfoti:
sounds good. i'll try monitoring with tensorboard and also digging around and doing some profiling. does edward have a reasonable way to save models and inference objects to disk? or were you just thinking pickle? thanks.

dustinvtran:
We use the recommended approach in TensorFlow, which is to [export/import `MetaGraph`s](https://www.tensorflow.org/programmers_guide/meta_graph) if you want to save the full graph to disk (i.e., both model and criticism so everything up to `inference.initialize()`). 

If you want to save just the model parameters for either the variational approximation or probability model, you can look into [`tf.Train.Saver`](https://www.tensorflow.org/programmers_guide/variables), which enables checkpointing and reusing models.

chmp:
Currently, I am also trying to use edward for time-series modeling and this PR aligns perfectly with my experiments. Thanks :).

So far I only tried the `example_ar-p` notebook and hope it's ok to offer some feedback:

- since `xrange` is python 2 only, it may be a good idea to replace it with `range`. As far as I can tell there shouldn't be any serious performance implications
- the MAP cell contains a missing new-line before `inference`

chmp:
I also experimented with using `tf.scan` to build the AR model. However, I ran into multiple problems, since scan generates recursive ops: `get_variables` needs to be modified to skip nodes already visited. Once this issue is fixed, `copy` dies to an infinite recursion (#306). For this issue, I have no fix yet.

dustinvtran:
Cool. Thanks for the update!

chmp:
wrt copy: at first glance, it does seem as if tf breaks down the while / scan ops into more fundamental control flow ops. I fear a more general solution is needed.

PR #485, opened 2017-02-28
Miscellaneous improvements to ed.evaluate and ed.ppc

dustinvtran:
A number of miscellaneous but fundamental improvements are made to Edward's criticism module.

1. `ed.evaluate` now does proper Monte Carlo for estimating posterior predictive means and (log) posterior predictive densities. To make this somewhat accurate, the default number of samples is now `n_samples=500`. In this move, I encourage using `copy()` to swap prior and posterior latent variables, but never to swap the prior with _just_ the posterior mean.

2. Both `ed.evaluate` and `ed.ppc` can work with implicit latent variables and implicit observed variables. This is done as part of the move to fully support implicit probabilistic models.

3. The key to the `latent_vars` argument in `ed.ppc` are now random variables and not strings.

4. I added unit tests.

### DIFF
--- a/docs/tex/tutorials/criticism.tex
+++ b/docs/tex/tutorials/criticism.tex
@@ -132,10 +132,7 @@ distribution.
 The \texttt{ed.ppc()} method provides a scaffold for studying
 various discrepancy functions.
 \begin{lstlisting}[language=Python]
-def T(xs, zs):
-  return tf.reduce_mean(xs[x_post])
-
-ed.ppc(T, data={x_post: x_train})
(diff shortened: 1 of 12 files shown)


rw:
I see that the `test_ppc.py` file executes the happy path for this code, which I assume is to protect against API signature regressions. Are there any meaningful predicates we could add that would turn these tests into ones that can pass or fail based on the result of a computation?

dustinvtran:
Yup, these are the simplest unit tests to get the momentum rolling. I think the next step would be to test the deterministic behavior. Unit tests with stochasticity can be a bit difficult to handle but are also possible. For example, @akucukelbir is writing tests that check if inference at least succeeds on toy problems. Similar ideas could be applied to check criticism.

PR #494, opened 2017-03-01
Add the docker

SnowMasaya:
I made a dockerfile for edward.

Reference

http://edwardlib.org/troubleshooting


dustinvtran:
Cool! Do you know how this can be checked? (I've never used Docker.)

pwl:
I can confirm that this works flawlessly with my setup (Arch Linux and nvidia graphics card with `nvidia-docker` installed).

Would it be possible to have an option to install the CPU version of TF instead of the GPU?

SnowMasaya:
> dustinvtran

When you check the it works well, you below the type command.(you move to the docker directory) 

build check

```
make build
```

test check in the docker environment(GPU)

```
make test
```

> pwl 

OK.
I will add the CPU version

SnowMasaya:
I already add the Docker file for CPU version.

If you use the it, you read the README.md

I wrote the how to use the Docker for CPU environment and GPU environment

pwl:
Thanks for the changes!  Out of curiosity, what is the reason to run `jupyter notebook` command in the Dockerfile instead of in the Makefile?

SnowMasaya:
@pwl 

>what is the reason to run jupyter notebook command in the Dockerfile instead of in the Makefile?

I think the most of the people would like to use "edward" in the notebook. 

When the people don't read the READEME, it is possible to use the notebook.

dustinvtran:
@SnowMasaya | Thanks again for this. Is there anything this is still waiting on? (@pwl can act as the code reviewer instead of me. :-)

pwl:
I can verify that both containers (CPU and GPU) work, I'd merge this.

PR #500, opened 2017-03-04
WIP: proper handling of recursive graphs

chmp:
When working with control flow constructs in tensorflow often graphs are defined recursively. Examples are `while_loop` and `scan`. Currently, such graphs are not properly handled in edward.

This PR addresses this issues. As of now, this branch of edwards allows to execute:

```python
minimal_op = tf.scan(lambda a, x: a + x, tf.random_normal([3]))
print(ed.copy(minimal_op).eval())
print(ed.get_variables(minimal_op))
```

Before merging, this branch requires proper tests. Further, there is a good chance that other functions (such as get_ancestors) need to be changed also.

Additional internal tensorflow API used:

- `Operation._add_input`
- `Operation._add_control_inputs`

### DIFF
--- a/edward/util/random_variables.py
+++ b/edward/util/random_variables.py
@@ -205,24 +205,6 @@ def copy(org_instance, dict_swap=None, scope="copied",
     else:
       new_original_op = None
 
-    # If it has control inputs, copy them.
-    new_control_inputs = []
-    for x in op.control_inputs:
-      elem = copy(x, dict_swap, scope, True, copy_q)
-      if not isinstance(elem, tf.Operation):
-        elem = tf.convert_to_tensor(elem)
-
-      new_control_inputs += [elem]
-
-    # If it has inputs, copy them.
-    new_inputs = []
-    for x in op.inputs:
-      elem = copy(x, dict_swap, scope, True, copy_q)
-      if not isinstance(elem, tf.Operation):
-        elem = tf.convert_to_tensor(elem)
-
-      new_inputs += [elem]
-
     # Make a copy of the node def.
     # As an instance of tensorflow.core.framework.graph_pb2.NodeDef, it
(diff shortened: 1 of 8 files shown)


edward/util/random_variables.py
```diff
@@ -579,9 +581,15 @@ def get_variables(x, collection=None):
...
   output = set([])
+  visited = set()
   nodes = set([x])
   while nodes:
     node = nodes.pop()
+
+    if node in visited:
+      continue
+    visited.add(node)
+
```
dustinvtran: A clear and simple fix. Should this also be done in `get_parents`, `get_children`, etc.?

Also I like that `set()` works over `set([])`. Could you replace the similar lines such as `output` to also do this?

chmp: Will do. `get_parents`, `get_children`, and the likes are on my list.

dustinvtran:
This looks great. Only one comment above. I'm surprised TensorFlow functionality exists to make adding those inputs so easy.

chmp:
I'm surprised, too. While I'm not too familiar with the tensorflow internals, it was quite interesting to see how it all works. Small caveat here: now edward is a bit more dependent on the internals of tensorflow. However, I did not come across any simple alternative.

Regarding your comment: I extended the recursion safety to all other relevant functions in `random_variables`. I also took the liberty to replace `set([x])` with the set literal `{x}`. All in the name of shaving off unnecessary characters ;).

chmp:
If you don't have any further comments, I'm happy with the current state of the code. I only would like check the tensorflow code once more to ensure that my changes to `copy` do not alter the semantics.

dustinvtran:
LGTM

chmp:
Perfect. Thanks.

PR #509, opened 2017-03-07
new example: hierarchical logistic regression edward

mariru:
Someone asked me how to implement logistic regression in Edward, so I added an example. This is a modification of `tf_hierarchical_logistic_regression.py` to use the Edward modeling language.


dustinvtran:
Thanks @mariru. The modification looks great. Travis is complaining about PEP8. (https://travis-ci.org/blei-lab/edward/jobs/208483743#L1196)

Issue #4, opened 2016-02-12
variance reduction techniques


dustinvtran:
- [x] use analytic form of q's entropy when available
- [x] use analytic form of KL(q(z) || p(z)) when available
- [ ] rao blackwellization
- [x] control variates (the vanilla one, nothing fancy yet)

RJT1990:
Antithetic variables could be another option if the approx distribution is symmetric?

dustinvtran:
Also a possibility! Antithetic variables are common in Monte Carlo, less so in variational inference. This means how it should be built, like control variates, should be general-purpose and not tied to a particular inference class.

bhargavvader:
This still relevant? Any tips/papers on how to get at this?

dustinvtran:
Apologies for the delay! Yes, it's very relevant. :) 

Here's my current thoughts on how to implement these:

1. __analytic KL.__ use the KL operator that some Google collaborators have been working on [[source]](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/distributions/python/ops/kullback_leibler.py) We can check if this operator is defined over pairs of random variables and apply it if it is.

2.  __rao blackwellization.__ this requires adding more variational inference infrastructure, so that we can update individual parameters in variational inference. (similar to how we manually update the parameters in the monte carlo classes)

3. __control variates.__ all easy to add. my suggestion would be to define utility functions outside the inference classes, and then call the function(s) during inference.

bhargavvader:
Thanks for the response! I'll look into this.

dustinvtran:
I have a WIP in the [`feature/rao-blackwellization`](https://github.com/blei-lab/edward/commits/feature/rao-blackwellization) branch. For a random variable `x_i`, it Rao-Blackwellizes the gradient with respect to other random variables in `latent_vars` and `data`. 

However, it doesn't Rao-Blackwellize with respect to the random variable's own `batch_shape`., i.e., `x_{-i}`. I'm not sure if there's enough information in `tf.contrib.distributions` to do this though. (For example, say we define `x = Normal(mu=tf.zeros(2), sigma=tf.ones(2))`. Can we tell that first element of `x` is parameterized by the first zero and one?

lbollar:
I was interested in taking on the control variates piece. Best to create the utility function in separate module inside inference directory then?

Thanks!

dustinvtran:
That sounds right. Then we can apply the utility function as a control variate for many methods.

[...]

PR #513, opened 2017-03-07
Added helper function to calculate optimal control variate coefficient

lbollar:
WIP 
Referencing [#4](https://github.com/blei-lab/edward/issues/4), this is an attempt to create a helper function for implementing a vanilla control variates method for variance reduction.

Right now, I just have this calculating the coefficient as this is what I thought was really needed to use this technique.

I tested this function with following implementation in build_score_loss_and_gradients in klqp.py: 
```
  f = q_log_prob * losses
  h = q_log_prob

  a = get_control_variate_scalar(f, h)

  grads = tf.gradients(
      -tf.reduce_mean(q_log_prob * tf.stop_gradient(losses - a)),
      [v._ref() for v in var_list])
```

Using this on the [Mixture Model](http://edwardlib.org/tutorials/unsupervised) example, this seemed to converge in under half the the time of original implementation and was achieving a loss of under 1000 in under 400 iterations. 

Apologies for latex errors that might exist, that is not my strong point.

Is this the right direction?


dustinvtran:
Neat. The implementation is clear. Moving it into `edward/util/tensorflow.py` seems like the best place for now until we build up more control variates. 

Can you also add a unit test? You can use similar tests such as [`test_reduce_logmeanexp.py`](https://github.com/blei-lab/edward/blob/master/tests/test-util/test_reduce_logmeanexp.py) as reference. To run tests, you'll need to install `py.test`. This lets you call, e.g., `py.test test_reduce_logmeanexp.py`.

Once we merge this, we can go ahead and start adding it as an option for the black box methods.

dustinvtran:
Great! Lastly, I think you need to add `get_control_variate_scalar` to `edward/__init__.py`, so we can call `ed.get_control_variate_scalar`.

dustinvtran:
Looks like there's a merge conflict. You'll have to merge/rebase to master; also I'd prefer it if it was added to __init__ in alphabetical order :)

lbollar:
Sorry, lack of attention to detail on my part :)
Now done.

PR #531, opened 2017-03-10
Add regularization loss to MAP

CTTC:
This PR adds regularization loss to `map.py`, which then can better support `weights_regularizer` and `biases_regularizer` in tf.slim


dustinvtran:
A simple and clear implementation—thanks. Can you add a test which uses it, building on top of [`test_map.py`](https://github.com/blei-lab/edward/blob/master/tests/test-inferences/test_map.py)?
 To run the unit tests, we use [py.test](http://doc.pytest.org/en/latest/): run `py.test tests/`. To run specific unit tests, call `py.test` individually, e.g., `py.test tests/test_metrics.py`.

 We should also update the docstring for the MAP class.

CTTC:
I add an example to illustrate how to add regularization loss via tf.slim. Since the example uses a quite small neural network, so the regularization effect is not very obvious. But by assigning the `weight_decay` to a large number, we can see the difference in the result.

dustinvtran:
Got it. Is there a way to add regularization without using Slim? Here's a simple unit test you can implement: define MLE on a normal likelihood, add the L2 penalty, and check that optimizing this loss is the same as MAP with a normal prior. Specifically:
```python
  def test_normal_losses_run(self):
    with self.test_session() as sess:
      x_data = np.array([0.0] * 50, dtype=np.float32)

      mu = tf.Variable(0.0)
      x = Normal(mu=tf.ones(50) * mu, sigma=1.0)
     
     # todo: add regularization, equivalent to Normal(0, 1) on mu

      inference = ed.MAP(data={x: x_data})
      inference.run(n_iter=1000)

      # todo: check that the answer is the same as in the MAP case with mu / qmu
```

Issue #502, opened 2017-03-04
Suggestion: Add a API reference page


chmp:
In its current form, the documentation of edward is outstanding, when read in order. However, when read out of order, it can be difficult to find the right page. I like to suggest adding an index / reference page, collecting links to all documented functions. 

This suggestion may have overlap with #496 and #497.

dustinvtran:
Thanks. I've been thinking hard about this problem. The difficulty is in managing two sidebars, both on the left for browsing the website, and on the top for browsing the API.

Maybe the solution is to replace the current sidebar with the top sidebar, for the API. They can still navigate to other aspects of the website via clicking the Edward logo or a big "Home" button. We can then add a top sidebar for even more fine-grained navigation.

chmp:
I have to admit, I wasn't as ambitious :). My idea was just to add a new page, which has a list of all all classes available, linking into other existing docs. Since edward is quite lean, it should be usable without further subdivisions. I started with it, but then prioritized looking into #525. If you want to I can open a PR with my unfinished draft.

It looks like that at the moment:

![edward_reference](https://cloud.githubusercontent.com/assets/949987/23821368/6993825c-0630-11e7-8741-ee058220e14b.png)

The links are working. The stylesheet is clearly not ;). As for the implementation overhead, it is minimal since sphinx is doing all the heavy lifting in cross-referencing the docs.

dustinvtran:
Cool! Mind sending a (work in progress) PR? I'd like to browse it a bit to see how it might be improved.

chmp:
Done. Looking over the code, it seems they are minimal indeed.

dustinvtran:
Added in #536.

PR #536, opened 2017-03-11
WIP: Add reference page

chmp:
Early implementation of #502.


dustinvtran:
Following #547, you can now do
```
\subsubsection{Model}

{{sphinx

}}

\subsubsection{Inference}

{{sphinx

}}

\subsubsection{Criticism}

{{sphinx

}}
```
which cleans up the styling a little bit.

chmp:
Nice. I adapted the sections accordingly. 

The list styling was an error on my part. While the sphinx docs only mention you need to indent the sublists, [this helpful stack overflow answer](http://stackoverflow.com/a/5553270) spelled out what to do in detail: one has to indent such that the new bullet lines up with the text of the previous level, here, the indent has to be 2 spaces.

I also extended the model list to include all models available in my venv. Maybe it makes sense to add a small caveat that the available models may differ depending on the tensorflow version.

@dustinvtran: could you check whether the lists are up-to date? Edward development is so rapid, that it's hard to keep up :).

It now looks like:

image

docs/tex/api/reference.tex
```diff
@@ -0,0 +1,93 @@
...
+  * :class:`edward.inferences.KLqp`
+
+    * :class:`edward.inferences.ReparameterizationKLqp`
+    * :class:`edward.inferences.ReparameterizationKLKLqp`
+    * :class:`edward.inferences.ReparameterizationEntropyKLqp`
+    * :class:`edward.inferences.ScoreKLqp`
+    * :class:`edward.inferences.ScoreKLKLqp`
+    * :class:`edward.inferences.ScoreEntropyKLqp`
+
+  * :class:`edward.inferences.ImplicitKLqp`
```
dustinvtran: ImplicitKLqp and WGANInference both inherit from GANInference

docs/tex/api/reference.tex
```diff
@@ -0,0 +1,93 @@
...
+  * :class:`edward.inferences.KLpq`
+  * Generative Adversial Network inferences:
+
+    * :class:`edward.inferences.GANInference`
+    * :class:`edward.inferences.WGANInference`
+
+  * :class:`edward.inferences.MAP`
+  * :class:`edward.inferences.Laplace`
```
dustinvtran: Laplace inherits from MAP

docs/tex/api/reference.tex
```diff
@@ -0,0 +1,93 @@
...
+
+\subsubsection{Criticism}
+
+{{sphinx
+
+* :func:`edward.criticisms.evaluate`
+* :func:`edward.criticisms.ppc`
+
+}}
+
```
dustinvtran: we try to remove all blanklines at the end of a file

docs/tex/api/reference.tex
```diff
@@ -0,0 +1,93 @@
...
+{{navbar}}
+
+\subsubsection{Models}
+
+{{sphinx
+
+* :class:`edward.models.RandomVariable`
+  
+  * :class:`edward.models.Bernoulli`
```
dustinvtran: Is there any way we can link these subclasses to TensorFlow distribution's API? Not a big priority, we can probably hack it when post-processing Sphinx blocks. I only wonder if Sphinx has a built-in solution.

chmp: Sphinx has something build in (intersphinx), if TF has a sphinx based docs. I will look into it. Otherwise, the URL format should not be too difficult to reverse engineer.

chmp:
I will fix the issues you brought up, once I'm back home.

docs/tex/api/reference.tex
```diff
@@ -0,0 +1,93 @@
...
+  * :class:`edward.models.WishartCholesky`
+  * :class:`edward.models.WishartFull`
+
+}}
+
+\subsubsection{Inference}
+
+{{sphinx
+
+* :class:`edward.inferences.VariationalInference`
```
dustinvtran: to add above this line:
```
* :class:`edward.inferences.Inference`
```
(No need to indent other classes within this one.)

docs/tex/api/reference.tex
chmp: I hacked it :).

chmp:
The injection of tensorflow version and links is a bit hacky, but not too complicated. Your comments are all addressed. Personally, I would like to replace `{{sphinx ... }}` by `{%sphinx ... %}` to enable embedded variables without having to resort to single braces. If you're up for it I would be happy to change the remaining docs.

docs/tex/api/reference.tex
```diff
@@ -0,0 +1,65 @@
+\title{Reference}
+
+{{navbar}}
+
+\subsubsection{Models}
+
+Most edward models are wrappers around tensorflow distrbutions. Therefore, the
```
dustinvtran: + `Most edward models` -> `Most Edward random variables`
+ capitalize Edward, TensorFlow;
+ `distrbutions` -> `distributions`
+ `Therefore, the` -> `Therefore the`

dustinvtran:
Nice work! Very cool that the list of TensorFlow distributions is auto-generated too. 

Planning to merge after the minor suggested changes above.

> Personally, I would like to replace {{sphinx ... }} by {%sphinx ... %} to enable embedded variables without having to resort to single braces. If you're up for it I would be happy to change the remaining docs.

Sure, that makes sense. Let's make it a separate PR.

docs/tex/api/reference.tex
dustinvtran: Oh, and similar to the flattening of `Inference` subclasses, I think we can have the random variable subclasses not scaffolded within `RandomVariable`.

chmp:
Addressed your comments.

Issue #224, opened 2016-08-16
add a test which runs all scripts in examples/ to check basic failure modes


dustinvtran:
This test shouldn't check that we recover good inferences or anything like that. It should simply check obvious failure modes, such as if the code immediately errors when you try to run it.

akucukelbir:
After chatting with @dustinvtran we decided that tests should be carefully designed and that we would eventually curate the list of examples to be useful to newcomers.

akucukelbir:
@dustinvtran do we still want to do this, based on some of our recent discussions?

dustinvtran:
I think it would be good to test the notebooks in `docs/notebooks`; also see https://github.com/GPflow/GPflow/issues/373 and GPFlow's implementation (https://github.com/GPflow/GPflow/pull/374). As in their implementation, we can give each notebook a timeout option to avoid very long testing with notebooks such as `gan.ipynb`.

For the example scripts, I'm not sure. Your call!

akucukelbir:
cool! looking into it now.

let's test the notebooks first. i'm not sure what'll happen to the examples in the near future.

reinheimer:
There is a problem exactly with the GAN example in Jupyter. The final lines of the 
[complete trace](https://github.com/blei-lab/edward/files/865496/gan.zip):
```
...
InternalError (see above for traceback): Blas SGEMM launch failed : a.shape=(128, 100), b.shape=(100, 128), m=128, n=128, k=100
	 [[Node: Gen/fully_connected/MatMul = MatMul[T=DT_FLOAT, transpose_a=false, transpose_b=false, _device="/job:localhost/replica:0/task:0/gpu:0"](Gen/Uniform/sample/Reshape, Gen/fully_connected/weights/read)]]
```
Searching online, I just found this error associated with the running of more than one instance at the same time, which is not my case. Also, using the terminal, the example runs just fine. Should I submit an issue for edward repository?

I'm using tensorflow-gpu (1.0.1) and edward (1.2.4) in a virtual environment (venv in Python 3.4.2).

dustinvtran:
Fixed in #542.

PR #542, opened 2017-03-13
Tests/notebooks

akucukelbir:
We'd like to test (most of) our shiny new notebooks.

This addresses: https://github.com/blei-lab/edward/issues/224

I'm basically replicating the great solution of: https://github.com/GPflow/GPflow/pull/374

TODO
- [x] make sure it passes on Python 2.7
- [x] debug any travis issues (if any)


tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+                             kernel_name=pythonkernel,
+                             interrupt_on_timeout=True)
+    lfiles = glob.glob(nbpath + "*.ipynb")
+    for notebook_filename in lfiles:
+      if(os.path.basename(notebook_filename) not in blacklist):
+        t = time.time()
+        self._execNotebook(ep, notebook_filename, nbpath)
+        print(notebook_filename, 'took %g seconds.' % (time.time() - t))
+
+
```
dustinvtran: all other unit tests have only one blankline here

tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+import glob
+import traceback
+import unittest
+import sys
+import time
+import os
+
+
+class TestNotebooks(unittest.TestCase):
+  def _execNotebook(self, ep, notebook_filename, nbpath):
```
dustinvtran: all other unit tests have a blank line between the first class method and the `class` line. also we use snake case, `test_notebooks` and `_exec_notebook`.

tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+      except CellExecutionError:
+        print('-' * 60)
+        traceback.print_exc(file=sys.stdout)
+        print('-' * 60)
+        self.assertTrue(False,
+                        'Error executing the notebook %s. See above for error.'
+                        % notebook_filename)
+
+  def test_all_notebooks(self):
+    ''' Test all notebooks except blacklist. '''
```
dustinvtran: we use `"""` for docstrings

tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+import nbformat
+from nbconvert.preprocessors import ExecutePreprocessor
+from nbconvert.preprocessors.execute import CellExecutionError
+
+import glob
+import traceback
+import unittest
+import sys
+import time
+import os
```
dustinvtran: can you alphabetize the import statements, then add a blank line, then add the `from` import statements? this mimicks all other python files

tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class TestNotebooks(unittest.TestCase):
```
dustinvtran: following the other unit tests, should we use `tf.test.TestCase` instead?

tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+        print('-' * 60)
+        traceback.print_exc(file=sys.stdout)
+        print('-' * 60)
+        self.assertTrue(False,
+                        'Error executing the notebook %s. See above for error.'
+                        % notebook_filename)
+
+  def test_all_notebooks(self):
+    ''' Test all notebooks except blacklist. '''
+    blacklist = ['gan.ipynb', 'iclr2017.ipynb']
```
dustinvtran: We can remove `iclr2017.ipynb` off the black list. It's not expensive and should run to verify all the syntax works.

tests/test-notebooks/test_notebooks.py
akucukelbir: sure thing.

tests/test-notebooks/test_notebooks.py
akucukelbir: i thought about this. we don't depend on any specific tensorflow functionality here (like having access to a session). this is why i kept it as a `unittest`

tests/test-notebooks/test_notebooks.py
akucukelbir: great catch

tests/test-notebooks/test_notebooks.py
akucukelbir: done

tests/test-notebooks/test_notebooks.py
akucukelbir: okay. i wasn't sure if this was a long-term notebook. if so, we may want to rename it to something a little bit more descriptive.

tests/test-notebooks/test_notebooks.py
akucukelbir: done

akucukelbir:
re: questions:

1. i was under the impression that we would be removing examples that are now represented in a notebook. if so, i think it would make more sense to keep a single copy of the data and to keep it near the notebooks. 
2. just so that i'm on the same page: are you asking whether we can do a pep8 check on the code cells in our notebooks?

dustinvtran:
1. It would be nice to have the data closer to the notebooks. At the same time, it would also be nice to have only one directory housing any external data.
2. Correct.

tests/test-notebooks/test_notebooks.py
```diff
@@ -0,0 +1,49 @@
...
+        print('-' * 60)
+        self.assertTrue(False,
+                        'Error executing the notebook %s. See above for error.'
+                        % notebook_filename)
+
+  def test_all_notebooks(self):
+    """ Test all notebooks except blacklist. """
+    blacklist = ['gan.ipynb']
+    pythonkernel = 'python' + str(sys.version_info[0])
+    nbpath = '../../docs/notebooks/'
```
dustinvtran: Do you know how this nbpath works? My impression was that travis calls pytest within the repo's director and not within `tests/test-notebooks/`, which is what assume this nbpath is referring to.

akucukelbir:
1. how do you suggest we restructure the directories?

i'm not sure why the notebooks are within `\docs\`, which to me only involves things relating to our website.

perhaps we want to port `docs\notebooks` into the `examples` directory and have the `data` folder there? 

perhaps we should rename `\docs` into `\website`?

2. i'm looking into it. might be more of a headache than what it's worth though, especially when we're writing/developing new notebooks. but let me be get back to you on this.

dustinvtran:
> i'm not sure why the notebooks are within` \docs\`, which to me only involves things relating to our website.

It's only in there for legacy reasons. I agree we should move it. 

I'd prefer to separate the `notebooks/` directory from the `examples/` directory. If we make `notebooks/` a top-level directory, then I'd be okay with having two `data/` subdirectories, one for `examples/` and one for `notebooks/`.

2. I think it's important to lay out good practices regardless of format. That said, it's not necessary for this PR though—we can look into it later as part of another PR.

akucukelbir:
1. ok let's take care of that in a separate PR.

2. this makes more sense to take care of here. are you open to running another command beyond `pytest` during testing? i think i can do this fairly easily with `flake8` for example

akucukelbir:
Ah I missed that. Sure. `pep8` also works. The main question is whether we're okay with call(s) to `pep8` in travis. (Right now we only call `pytest`)

dustinvtran:
It's worth thinking about how to simplify but not a priority. (For example, do we know how the pytest command currently checks PEP8 in all files? Is it possible to pipe the nbconversion into it as well?)

akucukelbir:
I looked into it briefly and it looked kinda complicated, mainly due to how `pytest` looks for files to test.

dustinvtran:
Bump. Let me know if I can help in any way.

PR #544, opened 2017-03-14
Fix linear mixed effects model tutorial

dustinvtran:
@timshell: After playing with the notebook, I noticed some irregularities with the way it was written. For example, we used a placeholder for `service` but not for any of the other training data. Also there was a minor bug when forming the test predictions.

I modified these and revised some of the text. Let me know what you think.


timshell:
I like this a lot better @dustinvtran, awesome work! I did notice the inconsistencies in the notebook but was hesitant to drastically change them. That being said, had I done so, they wouldn't be as clear as you made them out to be.

Issue #354, opened 2016-11-19
visualization of criticism techniques


dustinvtran:
references
+ [stan's bayesplot package](https://github.com/stan-dev/bayesplot)
+ [pymc3 plots](https://pymc-devs.github.io/pymc3/api.html#module-pymc3.plots)
+ [seaborn examples](http://seaborn.pydata.org/examples/)
+ [bokeh gallery](http://bokeh.pydata.org/en/latest/docs/gallery.html)
+ [scikit-learn examples](http://scikit-learn.org/stable/auto_examples/)

lbollar:
I have started some work on this. I think duplicating some of the base behavior in bayesplot would be a good idea. I have started with density plots of the posterior samples. Before I tidy up and write some docstrings, would it be best to create a separate plots folder for visualization related work?

dustinvtran:
Awesome! Up to you; it depends on how tied they are to existing modules or if they apply across many of the existing modules. For example, the criticism folder might incorporate the visualization related work if they are interrogating the models or inference.

PR #545, opened 2017-03-14
WIP: Implementation of two PPC plots

lbollar:
Referencing [#354](https://github.com/blei-lab/edward/issues/354)

Very rough WIP request. Right now, I don't have these two functions using edward methods. They are designed at the moment to use numpy arrays. I would image the workflow would look like using posterior.sample() or edward.ppc() to generate the needed data, and then passing this to the plotting functions. 

I messed with actually using edward to generate the data inside the plotting functions, but this seemed cleaner at the moment. 

[Here](https://gist.github.com/lbollar/3737d3f668882d2da050ed20017f6e6b) is a test notebook showing what these plots generate at the moment. These were meant to replicate some of the work being done in bayesplot. I just used random numbers to simulate samples in the first one as I was running a bit short on time tonight.

Thanks!

### DIFF
--- a/edward/criticisms/ppc_plots.py
+++ b/edward/criticisms/ppc_plots.py
@@ -27,7 +27,7 @@ def ppc_density_plot(y, y_rep):
   n = y_rep.shape[0]
 
   for i in range(n):
-    ax = sns.kdeplot(y_rep[i,:], color="maroon", alpha=0.2, linewidth=0.8)
+    ax = sns.kdeplot(y_rep[i, :], color="maroon", alpha=0.2, linewidth=0.8)
 
   y_line = sns.plt.Line2D([], [], color='maroon', label='y')
   y_rep_line = sns.plt.Line2D([], [], color='maroon', alpha=0.2, label='y_rep')
@@ -38,6 +38,7 @@ def ppc_density_plot(y, y_rep):
   ax.legend(handles, labels)
 
   return ax
+
 
 def ppc_stat_hist_plot(y_stats, yrep_stats, stat_name=None, **kwargs):
   """Creates histogram plot comparing data to samples from posterior.
--- a/edward/criticisms/ppc_plots.py
+++ b/edward/criticisms/ppc_plots.py
@@ -2,7 +2,10 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import seaborn as sns
+try:
+  import seaborn as sns
+except ImportError:
+  pass
 
 
 def ppc_density_plot(y, y_rep):
--- a/edward/criticisms/ppc_plots.py
+++ b/edward/criticisms/ppc_plots.py
@@ -9,14 +9,14 @@ except ImportError:
 
 
 def ppc_density_plot(y, y_rep):
-  """Creates 1D kernel density plot comparing data to samples from posterior.
+  """Create 1D kernel density plot comparing data to samples from posterior.
 
   Parameters
   ----------
   y : np.ndarray
-    A 1-D numpy array
+    A 1-D Numpy array.
   y_rep : np.ndarray
-    A 2-D numpy array where rows represent different samples from posterior
+    A 2-D Numpy array where rows represent different samples from posterior.
 
   Returns
   -------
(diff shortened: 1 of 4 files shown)


dustinvtran:
Nice work. Seaborn makes sense, as well as the current structure of function arguments. I think you're right that they should work on NumPy arrays, and that the user is expected to call these after calling `ed.ppc`.  In any case, it's easy to change either way.

dustinvtran:
Bump. How's progress on the plots? I'd be happy to get this merged in incrementally. Don't feel obligated to submit a massive slate of work at a time. For example, adding tests to these existing plots would be sufficient to merge now.

lbollar:
Sorry Dustin, I went down the NUTS road for a bit and got busy at work, but
I will try and write tests for these and submit. See anything that needs
polishing here?

edward/criticisms/ppc_plots.py
```diff
@@ -0,0 +1,67 @@
+import seaborn as sns
```
dustinvtran: You should add
```python
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function
```
to the top of any new file.

edward/criticisms/ppc_plots.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def ppc_density_plot(y, y_rep):
+  """
+  Cretes 1D kernel density plot comparing data to samples from posterior.
```
dustinvtran: typo; you could also collapse these two lines, e.g.,
```python
"""Creates 1D ...
```
Same goes for docstring below.

edward/criticisms/ppc_plots.py
```diff
@@ -0,0 +1,67 @@
...
+
+  Parameters
+  ----------
+  y : numpy array
```
dustinvtran: `numpy array` -> `np.ndarray`

edward/criticisms/ppc_plots.py
```diff
@@ -0,0 +1,67 @@
...
+
+  Parameters
+  ----------
+  y_stats : numpy array
+    A 1D numpy array
+  yrep_stats : float
+    Float representing statistic value of observed data
+  stat_name : string
+    Optional string value for including statistic name in legend
+  Keyword Arguments
```
dustinvtran: `Keyword Arguments` -> `**kwargs`, to stay consistent with other Edward docstrings

edward/criticisms/ppc_plots.py
```diff
@@ -0,0 +1,67 @@
...
+  """
+  Cretes histogram plot comparing data to samples from posterior.
+
+  Parameters
+  ----------
+  y_stats : numpy array
+    A 1D numpy array
+  yrep_stats : float
+    Float representing statistic value of observed data
+  stat_name : string
```
dustinvtran: `string, optional`

edward/criticisms/ppc_plots.py
```diff
@@ -0,0 +1,69 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import seaborn as sns
```
dustinvtran: This adds an install requirement to Edward. I prefer minimal requirements, so can you wrap this in a try/except as follows?
```python
try:
  import seaborn as sns
except ImportError:
  pass
```

lbollar:
Looking at Travis, do i need to do the same thing for matplotlib in tests?

tests/test-criticisms/test_ppc_plots.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import matplotlib
+matplotlib.use('Agg', warn=False)
```
lbollar: I wasn't sure how necessary this line is, but I didn't know much about testing plotting functions, and so I took a look at what PyMC3 had done and took this from there.

dustinvtran: Got it. I think it's not needed. The aim of the test is just to show that the function successfully terminates.

dustinvtran:
Good catch; following above comment, no, but you should definitely add seaborn to `.travis.yml`. For example, after line 32,
```python
  - pip install seaborn
```

lbollar:
Looks like build failed for 2.7, and it looks display related maybe. Think that could possibly be related to what pycharm had in their test code relating to choosing a dislay.

dustinvtran:
I was able to solve this problem. Not sure how to add commits to your PR; can you add a commit making these exact changes? https://github.com/blei-lab/edward/commit/8983355215b57e248c03afaa71b907a3b7ae57fc

lbollar:
Sorry I haven't been able to figure this out. Looks like the same error on 2.7 still. Did I miss something?

lbollar:
I saw the addition of a sleep command [here](http://stackoverflow.com/questions/35403127/testing-matplotlib-based-plots-in-travis-ci).

```yml
before_script: # configure a headless display to test plot generation
- "export DISPLAY=:99.0"
- "sh -e /etc/init.d/xvfb start"
- sleep 3 # give xvfb some time to start
```
Not sure if that would do anything to fix this or not...

dustinvtran:
It seems to work on my local branch? I'll merge anyways. It might be a permission problem; if not, I'll look into it again.

PR #557, opened 2017-03-17
Add PMF example

siddharth-agrawal:
This is a commit towards #22 adding an example for probabilistic matrix factorization. I have added a very basic example with a dense rating matrix. I was not sure of how the results could be visualized, do let me know if you have suggestions.

### DIFF
--- a/examples/probabilistic_matrix_factorization.py
+++ b/examples/probabilistic_matrix_factorization.py
@@ -26,13 +26,13 @@ V_true = np.random.randn(D, M).astype(np.float32)
 # DATA
 R_true = build_toy_dataset(U_true, V_true, N, M)
 I_train = get_indicators(N, M)
-I_test = 1-I_train
+I_test = 1 - I_train
 
 # MODEL
 I = tf.placeholder(tf.float32, [N, M])
 U = Normal(mu=tf.zeros([D, N]), sigma=tf.ones([D, N]))
 V = Normal(mu=tf.zeros([D, M]), sigma=tf.ones([D, M]))
-R = Normal(mu=tf.matmul(tf.transpose(U), V)*I, sigma=tf.ones([N, M]))
+R = Normal(mu=tf.matmul(tf.transpose(U), V) * I, sigma=tf.ones([N, M]))
 
 # INFERENCE
 qU = Normal(mu=tf.Variable(tf.random_normal([D, N])),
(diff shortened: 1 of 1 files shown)


dustinvtran:
Nice work. Yeah, test data as a whole new matrix is an unusual application. Typically, it's held-out entries inside the matrix, or held-out rows/columns.

One that might be cool and doesn't require test data would be a simple heatmap of the training data's matrix. Then compare it to the heatmap of the matrix of predictions from the model.

siddharth-agrawal:
@dustinvtran I have added testing on the held out entries and heatmap visualization. I have used matplotlib as I couldn't get seaborn working.

siddharth-agrawal:
@dustinvtran Is there a way to avoid style failures before pushing?

dustinvtran:
Try the [`pep8`](https://pypi.python.org/pypi/pep8) package. You can then run pep8 on your script
```bash
pep8 probabilistic_matrix_factorization.py
```

examples/probabilistic_matrix_factorization.py
```diff
@@ -0,0 +1,57 @@
+import edward as ed
```
dustinvtran: I recommend styling the header/imports similar to the other examples scripts. In particular, include `#!/usr/bin/env python`, then add a short docstring written below via `"""[text]"""`, then add `from __future___ ...`, and alphabetize your import statements.

examples/probabilistic_matrix_factorization.py
```diff
@@ -0,0 +1,57 @@
...
+  return ind
+
+
+N = 50  # number of users
+M = 60  # number of movies
+D = 3  # number of latent factors
+
+# true latent factors
+U_true = np.random.randn(D, N).astype(np.float32)
+V_true = np.random.randn(D, M).astype(np.float32)
```
dustinvtran: can you make make these true latent factors larger/smaller? it's hard to tell if inference worked because the prior distributions capture these factors pretty well already.

dustinvtran:
Great works! Just a few comments above. The heatmaps look nice.

examples/probabilistic_matrix_factorization.py
siddharth-agrawal: I tried `N=20` and `M=30`, but the training/test errors were large in that case. I assumed that that was because of lack of data to train on. Do you want me to do something else?

siddharth-agrawal: For `N=20` and `M=30`, the mean squared error comes out to be ~0.30. Whereas, for `N=500` and `M=600`, it comes out to be ~0.01. For the current setting, it is ~0.06.

dustinvtran: Ah, got it! This works.

dustinvtran:
Thanks again.

PR #574, opened 2017-03-22
Added ali.py

AlexLewandowski:
More or less a copy of gan_inference.py with encoder/decoder, as well as a different generator loss for the joint learning task.


dustinvtran:
Nice job. Some preliminary feedback:

+ Is there a way to inherit from `GANInference` instead of `VariationalInference`? This avoids having to redefine a lot of methods. This is what `WGANInference`, `ImplicitKLqp` and `AdversarialVB` (in #449) do, which could be useful references.
+ Comparing to `ImplicitKLqp` and other VAE-like algorithms we do such as in `KLqp`, is there a way to implement `ALI` without explicitly passing in encoder and decoder as functions? This may or may not be redundant if you already passed in the observed random variable `xf`, which is the output from the decoder, and analogously `zf` for the encoder.
+ Can you provide a working example using the algorithm, preferably on real data?
+ Unit tests are nice. You can look into building off one provided for other algorithms here: https://github.com/blei-lab/edward/tree/master/tests/test-inferences. More details re:unit testing in http://edwardlib.org/contributing.

AlexLewandowski:
I've incorporated your feedback. As for a working example, I have a Jupyter notebook with an MNIST generation and reconstruction samples that I will upload soon. I'll also look into unit testing.

AlexLewandowski:
I'm going to close this as I wasn't working in the proper branch. Will reopen with demonstration later.

Issue #564, opened 2017-03-20
DP has no state across calls to sample


JohnReid:
The newly implemented `DirichletProcess` does not hold state across calls to `sample()`. This is counterintuitive (to me at least). See https://gist.github.com/JohnReid/284b25d4fcd27a6d079c67e821744423
I was expecting the atoms to be the same across the two calls

PR #575, opened 2017-03-22
Memoise beta, the stick breaking proportions of a DP

JohnReid:
A `DirichletProcess` should retain its beta parameters across calls to sample. I've implemented this in the same way as @dustinvtran did for the theta parameters in issue #564. Also I simplified the code somewhat. Now the body and condition functions for the `tf.while_loop` are defined in `sample()` so that the scoping is easier. The theta and beta tensors are added to as needed and don't need to be sampled in `__init__`, rather they are initialised as empty. Also I ran the code through AutoPep8, I hope that is OK. I'm not sure what edward's standards are re. indentation.

### DIFF
--- a/edward/models/dirichlet_process.py
+++ b/edward/models/dirichlet_process.py
@@ -54,13 +54,16 @@ class DirichletProcess(RandomVariable, Distribution):
         # Create empty tensor to store future atoms.
         self._theta = tf.zeros(
             [0] +
-            self.get_batch_shape().as_list() +
-            self.get_event_shape().as_list())
+              self.get_batch_shape().as_list() +
+              self.get_event_shape().as_list(),
+            dtype=self._base.dtype)
 
         # Instantiate beta distribution for stick breaking proportions.
         self._betadist = Beta(a=tf.ones_like(self.alpha), b=self.alpha)
         # Create empty tensor to store stick breaking proportions.
-        self._beta = tf.zeros([0] + self.get_batch_shape().as_list())
+        self._beta = tf.zeros(
+            [0] + self.get_batch_shape().as_list(),
+            dtype=self._betadist.dtype)
 
         super(DirichletProcess, self).__init__(
             dtype=tf.int32,
@@ -147,7 +150,7 @@ class DirichletProcess(RandomVariable, Distribution):
     bools = tf.ones([n] + self.get_batch_shape().as_list(), dtype=tf.bool)
 
     # Initialize all samples as zero, they will be overwritten in any case
-    draws = tf.zeros([n] + batch_shape + event_shape)
+    draws = tf.zeros([n] + batch_shape + event_shape, dtype=self._base.dtype)
 
     # Calculate shape invariance conditions for theta and beta as these
     # can change shape between loop iterations.
--- a/edward/models/dirichlet_process.py
+++ b/edward/models/dirichlet_process.py
@@ -38,7 +38,7 @@ class DirichletProcess(RandomVariable, Distribution):
     >>>
     >>> # vector of concentration parameters, matrix of Exponentials
     >>> dp = DirichletProcess(tf.constant([0.1, 0.4]),
-        ...                       Exponential, lam=tf.ones([5, 3]))
+    ...                       Exponential, lam=tf.ones([5, 3]))
     >>> assert dp.get_shape() == (2, 5, 3)
     """
     parameters = locals()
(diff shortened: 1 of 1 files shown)


dustinvtran:
Thanks for this.

We use 2-indent, following [TensorFlow's style](https://www.tensorflow.org/community/style_guide). I wasn't able to read the diff since the 2-to-4 swap makes it difficult. One comment from your description:

> the body and condition functions for the tf.while_loop are defined in sample() so that the scoping is easier

I thought about this a little and think it's preferable to have them outside the function. While scoping is immediate with nested functions, it requires redefining the function for each call to sample, which can be an unnecessary expense.

dustinvtran:
In your branch, try:
```bash
git reset HEAD~1
```

dustinvtran:
I also recommend using [`pep8`](https://pypi.python.org/pypi/pep8) to manually check your file. I personally prefer the manual approach as it enforces good practices so eventually you always write PEP8-able code. :)

JohnReid:
OK @dustinvtran, I've changed the indentation back and changed the functions to be methods again.

edward/models/dirichlet_process.py
```diff
@@ -37,7 +38,7 @@ def __init__(self, alpha, base_cls, validate_args=False, allow_nan_stats=True,
     >>>
     >>> # vector of concentration parameters, matrix of Exponentials
     >>> dp = DirichletProcess(tf.constant([0.1, 0.4]),
-    ...                       Exponential, lam=tf.ones([5, 3]))
+        ...                       Exponential, lam=tf.ones([5, 3]))
```
dustinvtran: remove blankspaces

edward/models/dirichlet_process.py
```diff
@@ -10,8 +10,9 @@
 
 
 class DirichletProcess(RandomVariable, Distribution):
-  def __init__(self, alpha, base_cls, validate_args=False, allow_nan_stats=True,
-               name="DirichletProcess", value=None, *args, **kwargs):
+  def __init__(
```
dustinvtran: any reason you added a linebreak? i'd default to no break otherwise, to be consistent with other class definitions in edward

edward/models/dirichlet_process.py
```diff
@@ -51,18 +52,25 @@ def __init__(self, alpha, base_cls, validate_args=False, allow_nan_stats=True,
 
         # Instantiate base distribution.
         self._base = self._base_cls(*self._base_args, **self._base_kwargs)
-        # Define atoms of Dirichlet process, storing only the first as default.
-        self._theta = tf.expand_dims(
-            self._base.sample(self.get_batch_shape()), 0)
+        # Create empty tensor to store future atoms
```
dustinvtran: use period when ending comments that are sentences, similar to the comment two lines above

edward/models/dirichlet_process.py
```diff
@@ -51,18 +52,25 @@ def __init__(self, alpha, base_cls, validate_args=False, allow_nan_stats=True,
...
+            self.get_event_shape().as_list())
+
+        # Instantiate beta distribution for stick breaking proportions
+        self._betadist = Beta(a=tf.ones_like(self.alpha), b=self.alpha)
+        # Create empty tensor to store stick breaking proportions
+        self._beta = tf.zeros([0] + self.get_batch_shape().as_list())
 
         super(DirichletProcess, self).__init__(
-            dtype=tf.int32,
+            dtype=tf.float32,
```
dustinvtran: Could you comment on this change?

JohnReid: I could have the wrong end of the stick here but, isn't the process associated with the thetas? They don't have int32 type. Perhaps this should be `base_cls.dtype`

edward/models/dirichlet_process.py
JohnReid: will do

edward/models/dirichlet_process.py
JohnReid: Will do

edward/models/dirichlet_process.py
JohnReid: Flake8 told me the line was too long by 1 char

dustinvtran: Ah, got it. It should pass pep8 according to the additional rules we set up, so no need to bother (https://github.com/blei-lab/edward/blob/master/setup.cfg). (the `pep8` python package automatically uses additional rules from setup.cfg; not sure about flake8.)

edward/models/dirichlet_process.py
dustinvtran: The Dirichlet process always returns atoms, i.e., it is (almost surely) discrete even when the base distribution is continuous. This made me opt to use `tf.int32`. That said, [the jury's still out](https://github.com/tensorflow/tensorflow/issues/8585) on what dtype should be used for discrete distributions.

dustinvtran:
The test for [DP sampling](https://travis-ci.org/blei-lab/edward/jobs/213995443#L966) is failing. It seems to say that you're concatenating two tensors of differing dtype. I think this can be solved by passing in the dtype whenever you're first initializing a tensor to be a bunch of zeros. For example,
```python
self._theta = tf.zeros(..., dtype=self._base.dtype)
...
self._beta = tf.zeros(..., dtype=self._betadist.dtype)
...
draws = tf.zeros([n] + batch_shape + event_shape, dtype=self._base.dtype)
```

JohnReid:
Thanks for all the help. I think I managed to get it to pass the tests. I need to start using `pep8` more often I think. I hope it is useful.

PR #588, opened 2017-03-25
Feature/simpler conjugacy

matthewdhoffman:
This is a new, improved, rewritten version of the feature/conjugacy module I started months ago.

It mostly boils down to supporting one function: `ed.complete_conditional()`. This function takes as input `rv` (a `RandomVariable`) and `blanket` (a set of other `RandomVariable`s), and tries to do some exponential-family algebra to figure out the conditional distribution of `rv` given `blanket`. (Behavior is unchanged if `rv in blanket == True`.)

The other user-facing change is adding a `sample_shape` kwarg to `RandomVariable`, which brings Edward syntax a bit more up to date with the `tf.contrib.distribution` syntax and reduces the amount of algebra the system has to do (compared with syntax like `Normal(mu + tf.zeros([5]))`). This supersedes PR #323.

The high-level algorithm is:
1. Build a TF node that computes the full log-joint distribution of `rv` and all of the (other) `RandomVariable`s in `blanket`.
2. Crawl down the TF graph depth-first from the log-joint node until we hit either
  a. A member of `blanket` (which truncates the search) or
  b. `rv` or a nonlinear function of `rv` (which truncates the search and adds the node that stopped the search to a list of sufficient-statistic nodes).
3. Compute an s-expression-like intermediate representation of all sufficient-statistic nodes, and do some algebra on those nodes to simplify them (e.g., `log(mul(a, b))` becomes `add(log(a), log(b))`).
4. Look at the set of sufficient statistic nodes (and the support of `rv`) to see if there's a match in our table of exponential-family distributions—if not, we're done.
5. If there's a match, copy the log-joint node to a scratch namespace, replacing all sufficient-statistic nodes with placeholders.
6. Compute natural parameters in that scratch namespace by calling `tf.gradients()` with respect to the sufficient-statistic placeholders on the copied log-joint. This trick, which exploits the trivial identity ∂/∂t (η•t) = η, is there to save us from worrying about shapes.
7. Copy those natural parameters back to the original namespace, getting rid of any placeholders, and construct a new exponential-family `RandomVariable` with those natural parameters.


matthewdhoffman:
Oops, missed some PEP8 stuff and broken tests. I'll fix these.

dustinvtran:
Excellent work. I'm really excited about the practical implications for Edward, and also more researchy implications. For example, this could definitely be part of a NIPS paper aiming to bring back conjugacy in a world of automated/generic inference. (which we discussed almost two years ago and was hinted at in our ICLR paper. But now we have the tools go in-depth and validate its need.)

Comments:

1. Could you explain the copy/copying back in `complete_conditional`? I didn't get why you can't take gradients with respect to the original sufficient statistic nodes rather than placeholders.
2.  I like the design of `simplify`, where expressions are represented as nested tuples, and `simplify` modifies the nested structure according to a collection of functions. Is `as_float` casting necessary? Also, can we simplify (no pun intended) the implementation by avoiding a `NodeWrapper` class? Or does this complicate the code downstream?
3. `sample_shape` may have unintended consequences on existing algorithms since`Normal(0.0, 1.0, sample_shape=10)` represents a single random variable sampled 10 times, whereas `Normal(tf.zeros(10), tf.ones(10))` represents a batch of 10 random variables. Can you make a separate PR so we can merge this feature first and robustify it?
4. Can you also make a separate PR for `ParamMixture`?
5. Instead of passing in the blanket explicitly to `complete_conditional`, we should automate it via a `get_blanket` utility function; the argument can be optional. I'll make a separate PR adding this function.
6. The `support` attribute in random variables should be moved from `edward.inferences` into `edward.models`. I think they'll make it into `contrib.distributions` eventually. (That reminds me that I should draft the next spec for supports with Josh.) The same could be said of `conjugate_log_prob`, which is what you do for `ParamMixture`. Not sure about whether we should move `_suff_stat_to_dist`.

matthewdhoffman:
Thanks for the comments. I'm also excited to revisit the empirical question of what (if anything) conjugacy buys us over more black-box methods.

I'll pull out `sample_shape` and `ParamMixture` as separate PRs, make `blanket` optional, and move/refactor `support` and `conjugate_log_prob`. Regarding other questions:

1. The copy/copyback approach is annoying (and slow, because graph manipulation in TF is slow), and ideally I'd like to find a way to get rid of it. The problem it solves is that sometimes sufficient-statistic nodes can depend on other sufficient-statistic nodes. E.g., `log(x)` depends on `x`, and so taking the gradient w.r.t. `x` also gives you some unintended natural parameters. I considered using `stop_gradient()` in `conjugate_log_prob()`, but worried that that could have undesirable downstream consequences. It might be worth revisiting at some point, though.

2. Simplifying simplify.py: Good suggestions. I'll double check the implementation and decide whether `as_float()` and `NodeWrapper` are really necessary/helpful.

dustinvtran:
That makes sense. Too bad `tf.gradients` doesn't already let one specify the stop nodes rather than specify `tf.stop_gradient` when building the graph. I agree the alternatives sound like a significant undertaking.

Once you tease out `sample_shape`/`ParamMixture` I can add commits to those branches to help. I won't push any commits to this `feature/simpler_conjugacy` branch (yet).

edward/inferences/conjugacy/conjugacy.py
```diff
@@ -0,0 +1,246 @@
...
+    # nodes involving the sufficient statistic nodes are swapped for new
+    # nodes that depend linearly on the sufficient statistic placeholders.
+    s_stat_nodes = []
+    s_stat_replacements = []
+    s_stat_placeholders = []
+    swap_dict = {}
+    swap_back = {}
+    for s_stat in s_stat_exprs.keys():
+      s_stat_shape = s_stat_exprs[s_stat][0][0].get_shape()
+      s_stat_placeholder = tf.placeholder(np.float32, s_stat_shape)
```
dustinvtran: Is the float32 for placeholders because the log joint is float32?

edward/inferences/conjugacy/conjugacy.py
```diff
@@ -0,0 +1,246 @@
...
+_suff_stat_to_dist['real'][(('#CPow2.0000e+00', ('#x',)),
+                            ('#x',))] = (
+    rvs.Normal, normal_from_natural_params)
+
+
+def _log_joint_name(blanket):
+  return '_log_joint_of_' + ('&'.join([i.name[:-1] for i in blanket])) + '_'
+
+
+def get_log_joint(blanket):
```
matthewdhoffman: That's right. I introduced this caching for other reasons (which aren't all that relevant now), but it still speeds up graph construction.

But maybe it makes sense to move that caching a level up, so that each `RandomVariable` caches its `conjugate_log_prob()`?

(The term "cache" is a little funny here, since the graph remembers everything regardless. We're just caching a pointer.)

edward/inferences/conjugacy/conjugacy.py
```diff
@@ -0,0 +1,246 @@
...
+      terms.append(tf.reduce_sum(b.conjugate_log_prob()))
+    result = tf.add_n(terms, name=scope)
+    g.add_to_collection(blanket_name, result)
+    return result
+
+
+def complete_conditional(rv, blanket, log_joint=None):
+  with tf.name_scope('complete_conditional_%s' % rv.name) as scope:
+    # log_joint holds all the information we need to get a conditional.
+    extended_blanket = copy(blanket)
```
matthewdhoffman: Good catch; there's a bug here. I meant to add `rv` to `blanket` here. It's fixed now.

edward/inferences/conjugacy/conjugacy.py
matthewdhoffman: It's because we want to take the gradient w.r.t. the placeholders, and TF doesn't like taking gradients w.r.t. non-floats.

mariru:
@matthewdhoffman very much looking forward to this feature! Checked out the branch to play around with the examples. Direct import wasn't working so I added a line to the `__init__.py` file

dustinvtran:
Is there anything this PR is still waiting on? Happy to merge now. I have some minor suggestions, but I'll submit another PR after this one.

matthewdhoffman:
Nope, I think it's at a good checkpoint—let's merge. I haven't done the `conjugate_log_prob()` refactoring yet, but I probably won't have much time to look at it in the next couple of days and it can happen separately.

I made `blanket` optional, but added a warning about using that feature multiple times. (It'll throw a gnarly error if the user tries to do that—probably we could add a warning check.)

Also, it turns out that `get_blanket()` doesn't really do what we need here, since it only looks at the Markov blanket of the directed model, not the moralized graph. It might be nice at some point to add a moralization routine or Bayes ball or something to figure out what we actually need to condition on, but it shouldn't have much of a performance impact except possibly on graph creation.

dawenl:
@matthewdhoffman you will gradually make all of our jobs meaningless with this PR :)

PR #593, opened 2017-03-29
Fix broken reading-data link in data.tex

rdipietro:
The three-methods-of-getting-data TensorFlow link is broken. This just updates the link.


dustinvtran:
Thanks for the catch. Can you also fix the broken link on line 34 of the same file?

rdipietro:
Sorry about that – I quickly checked for other occurrences but missed that. All set.

PR #597, opened 2017-04-03
Feature/ali

AlexLewandowski:
Following the pull request from last week, I added an example for ali.py. The MNIST reconstructions are not perfect, but I haven't fiddled with the hyperparameters very much.

### DIFF
--- a/edward/inferences/ali.py
+++ b/edward/inferences/ali.py
@@ -1,0 +1,86 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import six
+import tensorflow as tf
+
+from edward.inferences.gan_inference import GANInference
+from edward.util import get_session
+
+
+class ALI(GANInference):
+  """Adversarially Learned Inference (Dumoulin et al., 2016) or
+  Bidirectional Generative Adversarial Networks (Donahue et al., 2016)
+  for joint learning of generator and inference networks.
+
+  Works for the class of implicit (and differentiable) probabilistic
+  models. These models do not require a tractable density and assume
+  only a program that generates samples.
+  """
+  def __init__(self, *args, **kwargs):
+    """
+    Notes
+    -----
+    ``ALI`` matches a mapping from data to latent variables and a
+    mapping from latent variables to data through a joint
+    discriminator. The encoder approximates the posterior p(z|x)
+    when the network is stochastic.
+
+    In building the computation graph for inference, the
+    discriminator's parameters can be accessed with the variable scope
+    "Disc".
+    In building the computation graph for inference, the
+    encoder and decoder parameters can be accessed with the variable scope
+    "Gen".
+
+    Examples
+    --------
+    >>> with tf.variable_scope("Gen"):
+    >>>   xf = gen_data(z_ph)
+    >>>   zf = gen_latent(x_ph)
+    >>> inference = ed.ALI({xf: x_data, zf: z_samples}, discriminator)
+    """
+    super(ALI, self).__init__(*args, **kwargs)
+
+  def initialize(self, *args, **kwargs):
+    super(ALI, self).initialize(*args, **kwargs)
+
+  def build_loss_and_gradients(self, var_list):
+    x_true = list(six.itervalues(self.data))[0]
+    x_fake = list(six.iterkeys(self.data))[0]
+    z_true = list(six.itervalues(self.data))[1]
+    z_fake = list(six.iterkeys(self.data))[1]
+    with tf.variable_scope("Disc"):
+        # xfzt := x_fake, z_true
+        d_xtzf = self.discriminator(x_true, z_fake)
+    with tf.variable_scope("Disc", reuse=True):
+        # xfzt := x_fake, z_true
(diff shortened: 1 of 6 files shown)


dustinvtran:
Nice work! Quick comment from a glimpse (will add more details later): since you're only calling parent methods for `update` and `initialize`, you could probably remove those implementations from `ALI`.

edward/inferences/ali.py
```diff
@@ -0,0 +1,86 @@
...
+  only a program that generates samples.
+  """
+  def __init__(self, *args, **kwargs):
+    """
+    Notes
+    -----
+    ``ALI`` matches a mapping from data to latent variables and a
+    mapping from latent variables to data through a joint
+    discriminator. The encoder approximates the posterior p(z|x)
+    when the network is stochastic.
```
dustinvtran: Actually this is untrue. The terminology "Adversarially learned inference" is a misnomer because the distribution p( z | x) in GANs is degenerate. More accurately would be to say
```
Using a joint discriminator, ``ALI`` learns a mapping from noise to data 
as well as an inverse mapping from data to noise. The inference network 
is the inverse mapping.
```
We discuss this misconception  in [my paper](https://arxiv.org/abs/1702.08896) on deep implicit models (see the noise vs latent variables part of section 2.1).

I personally prefer the other paper which calls it "adversarial feature learning"/BiGANs; to me this description is more accurate. So I'd be for `BiGANInference`, but if you feel strongly about `ALI`, that's okay too.

edward/inferences/ali.py
```diff
@@ -0,0 +1,86 @@
...
+
+import six
+import tensorflow as tf
+
+from edward.inferences.gan_inference import GANInference
+from edward.util import get_session
+
+
+class ALI(GANInference):
+  """Adversarially Learned Inference (Dumoulin et al., 2016) or
```
dustinvtran: Both were published at ICLR 2017, so you can update the citation with ... et al., 2017.

edward/inferences/ali.py
```diff
@@ -0,0 +1,86 @@
...
+    In building the computation graph for inference, the
+    encoder and decoder parameters can be accessed with the variable scope
+    "Gen".
+
+    Examples
+    --------
+    >>> with tf.variable_scope("Gen"):
+    >>>   xf = gen_data(z_ph)
+    >>>   zf = gen_latent(x_ph)
+    >>> inference = ed.ALI({xf: x_data, zf: z_samples}, discriminator)
```
dustinvtran: To match the interface of other inference methods, what about
```python
ed.ALI(latent_vars, data, discriminative_network)  # signature
ed.ALI({zf: qz}, {xf: x_data}, discriminator)  # example
```
Here the model's noise `zf` is binded to the inference network output `qz`, and the model's generated output `xf` is binded to data `x_data`.

dustinvtran:
@AlexLewandowski Great work on the updates. Is this waiting on anything? Happy to merge.

AlexLewandowski:
Yeah, you can merge. I'm working on a project so if I have any other ideas, I'll push them this way.

edward/inferences/bigan_inference.py
```diff
@@ -0,0 +1,84 @@
...
+    In building the computation graph for inference, the
+    encoder and decoder parameters can be accessed with the variable scope
+    "Gen".
+
+    Examples
+    --------
+    >>> with tf.variable_scope("Gen"):
+    >>>   xf = gen_data(z_ph)
+    >>>   zf = gen_latent(x_ph)
+    >>> inference = ed.BiGANInference({latent_vars: qz}, {xf: x_data}, discriminator)
```
dustinvtran: Should this line be `{z_ph: zf}` and `{xf: x_ph}`? Trying to map notation to the lines above.

edward/inferences/bigan_inference.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    joint = latent_vars.copy()
+    joint.update(data)
+    
+    super(BiGANInference, self).__init__(joint, discriminator)
+
+  def build_loss_and_gradients(self, var_list):
+    
+    x_true = list(six.itervalues(self.data))[0]
```
dustinvtran: With a dictionary, `self.data` doesn't guarantee the first key-value pair is `x` and the second is `z`. I think you can keep the `self.latent_vars` / `self.data` attributes stored separately, doing something like this for `__init__`:
```python
    if not callable(discriminator):
      raise TypeError("discriminator must be a callable function.")

    self.discriminator = discriminator
    # call grandparent's method; avoid parent (GANInference)
    super(GANInference, self).__init__(latent_vars, data)
```
In addition, for consistency with other inferences, the `z`s should be ordered as `z_true: z_fake`, i.e., the prior latent variable is binded to the output of the inference network. With the above you can do
```python
    z_true = list(six.iterkeys(self.latent_vars))[0]
    z_fake = list(six.itervalues(self.latent_vars))[0]
```

AlexLewandowski: Yeah that's my mistake, I didn't realize that the order was different. I made the changes, hopefully this is more consistent.

PR #619, opened 2017-04-19
Update probabilistic_pca_subsampling.py

Phdntom:
`inference` not defined in batch print statement.

 Looks like `inference` --> `inference_w` is the appropriate fix.


dustinvtran:
LGTM

PR #622, opened 2017-04-26
Polished util/tensorflow.py

KonstantinLukaschenko:
Removed unused variables and imports from `util/tensorflow.py`


dustinvtran:
LGTM

PR #630, opened 2017-05-01
Notebooks/data feed example

janekberger:
Added a simple example for feeding batches to a placeholder to the bayesian linear regression notebook.


dustinvtran:
This is great! The modified code makes sense. Can you also rewrite some of the text to emphasize that this notebook is to teach how to do batch training? You can also change the title to make this emphasis to something like "Batch training" or "Data subsampling".

I can help later to reformat the notebook so it's also renderable on the website.

janekberger:
Thanks for the suggestions, I committed an update.

dustinvtran:
Apologies for the delay. I will merge this now. Then I'll add an additional commit for an equivalent `tex` version to be available on the website. Thanks again.

PR #632, opened 2017-05-01
* Fix undefined variable in bayesian neural network example.

KonstantinLukaschenko:
The tensorflow placeholder is defined as X (uppercase) in this example.


dustinvtran:
LGTM

PR #633, opened 2017-05-03
Unify number of iterations in the getting started notebook.

KonstantinLukaschenko:
The posterior plot title declares 1000 iterations, while the code uses 500. Unify the amount of iterations in this example to 1000 to avoid confusion.


dustinvtran:
LGTM

Issue #636, opened 2017-05-05
You must feed a value for placeholder tensor when debug = True


synikitin:
Hello,

I have been trying to experiment with amortized inference, but binding a placeholder to an observable variable keeps crashing. I have tensorflow 1.1.0 and edward 1.3.1. Here is an example code:

```python
import tensorflow as tf
import edward as ed
import numpy as np

## data
n = 3
y = np.linspace(-1, 1, n)
y = y.reshape(3, 1)

## model
# generative
coef_gen = tf.Variable(tf.random_normal([1, 1]))
x_latent = ed.models.Normal(tf.zeros([n, 1]), tf.ones([n, 1]), name = "x_latent")
y_obs = ed.models.Normal(tf.multiply(coef_gen, x_latent), tf.ones(1), name = "y_obs")

# variational
coef_inf = tf.Variable(tf.random_normal([1, 1]))
y_ph = tf.placeholder(tf.float32, [n, 1])
x_inf = ed.models.Normal(tf.multiply(coef_inf, y_ph), tf.ones([n, 1]))

## inference
inference = ed.KLqp({x_latent : x_inf}, data = {y_obs : y_ph})
optimizer = tf.train.RMSPropOptimizer(0.01, epsilon = 1.0)
inference.initialize(optimizer = optimizer, n_iter = 500, n_samples = 5, debug = True)
init = tf.global_variables_initializer()
init.run()
inference.update({y_ph : y})
```

This can be avoided with `y_ph = tf.placeholder_with_default(tf.cast(y, tf.float32), [n, 1])`, but clearly defeats the purpose of feeding. 

Thanks for your help!

meta-inf:
@dustinvtran Seems you should feed feed_dict here:
https://github.com/blei-lab/edward/blob/master/edward/inferences/inference.py#L234

meta-inf:
Actually any code with inference.run(debug=True) crashes. (At least I tried the SGHMC Bayesian LR example and something of my own.)

meta-inf:
The problem is we did not feed the `data` dict when running `op_check`. A fix is referred above. 

I modified `feed_dict` in `Inference.update` as it did not handle cases like `data={rv: python_value}`, but I don't know if it breaks other things (I'm a bit lost in the codebase). @dustinvtran could you look at it?

dustinvtran:
Ah, got it. That makes sense re:`op_check` with feed dict. It does not break other things. A PR is welcome.

[...]

PR #642, opened 2017-05-16
Fixing #636

meta-inf:
...by fixing `feed_dict` and feed it to `op_check` in `Inference.update` and its overloaded instances.


dustinvtran:
Apologies for the delay. This PR looks good. I think we can solve the op_check and MonteCarlo issues without resorting to a `init_const_bindings`. I'll merge this now and make a separate commit.

PR #644, opened 2017-05-23
Update parameter name to Bernoulli()

fritzo:
I'm surprised doctests didn't catch this


dustinvtran:
Good catch. The unit tests don't run any of the docstrings (maybe they should?)

Issue #262, opened 2016-09-28
binary accuracy


mariru:
`ed.eval('binary_accuracy',[...])` evaluates to 1.0 when the Bernoulli mean is 0.5 irrespective of label. This can be missleading. Results (in terms of binary accuracy) seem great while your params are simply very close to 0.

dustinvtran:
That's interesting. So this is caused by `tf.round()` in `binary_accuracy()`, which always rounds 0.5 up to `1.0`. Should it round up or down randomly instead?

nitishjoshi25:
i was looking into this problem. `tf.round()` rounds `0.5` to `0` instead of rounding up to `1.0`.

dustinvtran:
Right. I think the easiest solution is to to do a random round if equal to 0.5. (using a tensorflow control flow op)

dustinvtran:
Fixed in #651.

PR #648, opened 2017-05-25
Changed class assignment from always 0 to random in "ed.eval('binary_accuracy',...)" when probability is 0.5 for either class

nitishjoshi25:
This is in reference to the discussion on [this](https://github.com/blei-lab/edward/issues/262) page.


nitishjoshi25:
One of the test is failing , and i cant figure out why.

Issue #464, opened 2017-02-15
Edward Roadmap


dustinvtran:
This is an outline of Edward going forward, at the least for Spring 2017.  Of course, comments are always welcome. I'm happy to change priorities subject to interest. Related issues are added in parentheses.

+ More community support
  + Mailing list
  + Pre-trained models in the model zoo (#28)
  + Repo for sharing Edward content such as papers, talks, and posters
  + Jupyter notebooks for all tutorials (#35)
  + Drafting stages on a fully open source handbook
+ More robustification
  + Integration tests, inference tests (#174)
  + Improved errors and warning messages
  + __Many__ minor bug fixes
+ More automation
  + Automated choice of posterior approximation
  + Automated transformations (#439)
  + Control variates (#4)
  + Convergence diagnostics (#5)
  + Automated hyperparameter tuning
+ More applications & tutorials
  + Sigmoid belief network with Gumbel-Softmax/Concrete distribution ([Jang et al., 2017](https://arxiv.org/abs/1611.01144); [Maddison et al., 2017](https://arxiv.org/abs/1611.00712)) (#299)
  + Large-scale LDA using (vanilla) stochastic variational inference [(Hoffman et al., 2013)](https://arxiv.org/abs/1206.7051) (#423, #463)
  + Generative model applied to ImageNet
  + Hidden Markov model (#450)
  + Other large-scale examples & tutorials (#22, #137)
  + More inference & criticism tutorials (#35)
  + More benchmarks, including a JMLR paper and dedicated webpage
+ More algorithms
  + More GAN algorithms, and in general for implicit probabilistic models (#444)
  + Conjugacy
    + (vanilla) stochastic variational inference [(Hoffman et al., 2013)](https://arxiv.org/abs/1206.7051)
    + Gibbs sampling [(Gelfand and Smith, 1990)](http://www.tandfonline.com/doi/abs/10.1080/01621459.1990.10476213)
  + Expressive variational families ([Rezende and Mohamed, 2015](https://arxiv.org/abs/1505.05770); [Ranganath et al., 2016](https://arxiv.org/abs/1511.02386)) (#7, #284, #447)
  + Gradient-based marginal optimization (Tran et al., 2017)
  + Operator variational inference [(Ranganath et al., 2016)](https://arxiv.org/abs/1610.09033)
  + Inference support for missing data & causality (#399)

[...]

Issue #353, opened 2016-11-19
visualization of inference/convergence diagnostics


dustinvtran:
for example, this includes:

+ trace plots for MCMC
+ loss function values by iteration for variational inference
+ parameter estimates or a predicted quantity by iteration

tensorboard is perfect for tracking these. all should live in tensorboard.

references
+ [stan's bayesplot package](https://github.com/stan-dev/bayesplot)
+ [pymc3 plots](https://pymc-devs.github.io/pymc3/api.html#module-pymc3.plots)
+ [seaborn examples](http://seaborn.pydata.org/examples/)
+ [bokeh gallery](http://bokeh.pydata.org/en/latest/docs/gallery.html)
+ [scikit-learn examples](http://scikit-learn.org/stable/auto_examples/)

dustinvtran:
#654 gets us 90% here. The last step is MCMC features. Namely, it would be nice to display essential MCMC plots such as a trace plot over the first X many dimensions of a parameter (and as usual, `tf.summary.histogram` to see all dimensions at once).

dustinvtran:
Closing this and adding a new issue for MCMC + TensorBoard.

Issue #351, opened 2016-11-19
model visualization


dustinvtran:
this is immediate from looking at the tensorboard. perhaps we can have a utility function which exports a picture of it, with various options such as visualizing only stochastic nodes vs visualizing both stochastic and deterministic nodes.

references
+ [model visualization in keras](https://keras.io/visualization/)

Issue #190, opened 2016-07-27
tf.name_scope()'s and tensorboard


dustinvtran:
We need to include these for [easy viewing of the graph](https://www.tensorflow.org/versions/master/how_tos/graph_viz/index.html) in Tensorboard. Also related to default diagnostics that can be visualized in Tensorboard, c.f., https://github.com/blei-lab/edward/issues/106.

matthewdhoffman:
It seems like the new random variable classes would benefit from this—the old `DistributionTensor`s encapsulated the parameters and samples in the same namespace.

dustinvtran:
One question is when to provide matplotlib/seaborn features for visualization, versus when to provide TensorBoard features for visualization.

+ My initial guess is that TensorBoard is useful for visualizing internals which users don't have explicit access to. This means #351 (visualizing a model's graph structure) and #353 (inference/convergence diagnostics).

+ On the other hand, matplotlib/seaborn is useful for visualizing things the users must call explicitly. This means #279 (visualizing high-dimensional distributions), #352 (visualizing fitted models), and #354 (criticizing fitted models).

(But it's a wash. The [TensorBoard embedding visualizer](https://www.tensorflow.org/get_started/embedding_viz) shows that it can also be an excellent platform for interactively visualizing fitted models.)

closedLoop:
I am working on this now and have made good progress.   Will report back shortly with a pull request

PR #653, opened 2017-05-27
TensorBoard Integration with Notebook Example

closedLoop:
Integration of TensorBoard with Notebook Example

@dustinvtran here's a quick demo of the changes
https://www.useloom.com/share/29e087b17cfc47b9b60662bff7843983

Contents:
 * Modification of `inference/inference.py` to more completely initialize Tensorboard
 * A Notebook with instructions and best practices to make your TensorBoard clean and useful

Main Features:
 * Automatically tracks all named latent variables for your given inference
   * This can be disabled or modified by passing in a custom list via `logvars`
 * Works for all inference classes.
 * Automatic run versioning to keep your logs clean and comparable in TensorBoard
 * Instructions for people to easily configure and launch their model with TensorBoard

Issues addressed
 * #464 satisfies part of the 'More TensorBoard support'
 * #190 the Notebook shows best practices for TensorBoard to look clean using Edward
 * #351 Partially addresses this by automatically visualizing all named latent variables
 * #353 This shows parameter MCMC values, but not loss functions.

Not addressed:
 * Storing inference specific loss and convergence information.  I saw that other were working on this and so I didn't touch it.  #598


edward/inferences/inference.py
```diff
@@ -169,11 +171,25 @@ def initialize(self, n_iter=1000, n_print=None, scale=None, logdir=None,
     logdir : str, optional
       Directory where event file will be written. For details,
       see ``tf.summary.FileWriter``. Default is to write nothing.
+    logrun : str, optional
```
dustinvtran: Could you describe your reasoning for splitting logdir vs logrun? I think I see it, although I wonder if logdir could just be made more powerful by letting you include a custom subdirectory naming. If unspecified, it would default to a subdirecty with the current UTC timestamp.

edward/inferences/inference.py
```diff
@@ -194,6 +210,16 @@ def initialize(self, n_iter=1000, n_print=None, scale=None, logdir=None,
 
     if logdir is not None:
       self.logging = True
+
+      if logrun is None:
+        # Set default to timestamp
+        logrun = datetime.strftime(datetime.utcnow(), "%Y%m%dT%H%M%S")
```
dustinvtran: Good standard! What's the T? 

Also, do you think adding `_` inbetween would make the directory name too long?

edward/inferences/inference.py
```diff
@@ -258,3 +284,48 @@ def finalize(self):
     """
     if self.logging:
       self.train_writer.close()
+
+  def set_log_variables(self, logvars=None, log_max_scalers_per_var=None):
+    """Logs variables to TensorBoard
```
dustinvtran: `Tensorboard.`

edward/inferences/inference.py
```diff
@@ -258,3 +284,48 @@ def finalize(self):
...
+
+     For each variable in logvars, creates 'scalar' and / or 'histogram' by
+     calling `tf.summary.scalar` or `tf.summary.histogram`
```
dustinvtran: To format code, wrap double single ticks, i.e., `\``tf.summary.scalar``\` (without the `\`). Also needs a period at end of sentence.

edward/inferences/inference.py
```diff
@@ -258,3 +284,48 @@ def finalize(self):
...
+
+     Returns
+     -------
+     None
+
+    """
+    if logvars is None:
+      logvars = []
+      for k in self.latent_vars:
+        logvars += get_variables(self.latent_vars[k])
```
dustinvtran: variables can also exist as model parameters. I recommend an additional loop applying this to keys of `self.data` and the priors (which are keys of `self.latent_vars`).

edward/version.py
```diff
@@ -1 +1 @@
-__version__ = '1.3.1'
+__version__ = '1.3.1-closedloop'
```
dustinvtran: Can remove this :)

examples/normal_sghmc.py
```diff
@@ -50,6 +50,7 @@ def mvn_plot_contours(z, label=False, ax=None):
 qz = Empirical(params=tf.Variable(tf.random_normal([5000, 2])))
 
 inference = ed.SGHMC({z: qz})
+inference.initialize(logdir='C:\\Users\skruz\\temp\\tensorflow\\')
```
dustinvtran: You can remove this argument to be more generic, say, `logdir='log'`.

dustinvtran:
Wow, this is an amazing PR @closedLoop. It's a very valuable contribution in making Edward fits easier to explore and also to diagnose model/learning problems. And thanks for the demo!

I added comments above. Two others

+ Can you also remove changes from the various notebooks unrelated to this PR?
+ I recommend displaying the TensorBoard pics in the tutorial. You can do that by using
```python
![GAN Fig 0](https://raw.githubusercontent.com/blei-lab/edward/master/docs/images/gan-fig0.png)
```
Then you can explain them as you do well in the video demo.

edward/inferences/inference.py
dustinvtran: Also: if logrun can easily be specified as part of logdir that would also enable the timestampping for those who prefer to control inference using `initialize()` then `updates()`. (that is, they don't call `run()`)

edward/inferences/inference.py
```diff
@@ -258,3 +284,48 @@ def finalize(self):
...
+      # If variable is a one dimensional tensor, log each element in the tensor
+      # individually. Only log the first log_max_scalers_per_var variables
+      if len(var.shape) == 1:
+        for i in range(var.shape[0]):
+          if log_max_scalers_per_var is None or i < log_max_scalers_per_var:
+            tf.summary.scalar('{}/{}'.format(var_name, i), var[i])
+
+      # If var is multi-dimensional, log the distribution
+      if len(var.shape) > 0 and np.max(var.shape) > 1:
+        tf.summary.histogram(var_name, var)
```
dustinvtran: Some of this conflicts with the current tensorboard tracking in `variational_inference.py` (see #654). My guess is you can remove the parameter histograms in `variational_inference.py`. 

I also recommend checking to see how the naming systems match.

edward/inferences/inference.py
closedLoop: They way I thought about it is that `logrun` will tend to be pretty static for a given project and you want a way to cleanly separate info each time you refit your models.  Therefor you can leave it blank as you arbitrarily refine the model and use the timestamps to track your progress.  Else you could set `logrun` to something like `using_adam` vs `using_adagrad` to highlight in TensorBoard the convergence results of each.

I wanted to keep `logdir` a string since it is a string in TensorBoard, and given it has the same name I wanted to make the usage similar to avoid confusion.

edward/inferences/inference.py
closedLoop: The T was a means of just separating the date and the time segments like most of the conventions here [https://www.loggly.com/docs/automated-parsing/#system](https://www.loggly.com/docs/automated-parsing/#system) and also removing all of the other non-numeric characters from the conventions.  However after using this in tensorboard a bit.  I'm going to replace the 'T' with a '_' just to keep it easier to read on the eye.

edward/inferences/inference.py
dustinvtran: Users can already separate info each time they refit a model by tweaking the `logdir` argument.  A typical use case would be
```python
inference.initialize(logdir='log/lr_{}_nsamples_{}'.format(learning_rate_stepsize, n_samples))
```
With `logdir`, users have full flexibility in setting however many subdirectories and filenames they want. `logdir + logrun` sort of imposes structure, where a user can specify `logrun` already inside `logdir`.

That said, I love the idea of having a default subdirectory inside `logdir` which timestamps the run. It seems impossible to turn it on/off without having an additional argument. So then maybe `logrun` should just be a boolean, and its name+default be changed to something like `log_timestamp=True`?

closedLoop:
@dustinvtran this has been cleaned up integrated with the gradient and loss tracking from the merged pull request #598 .  I think it's looking really good and I've tested it on some larger models and it works well there too.

Main changes:
 * parameter naming is consistent with logging in ``variational_inference.py``
 * ``logrun`` has been changed to ``log_timestamp`` boolean based on your suggestion.
 * I've removed ``log_max_scalars_per_var`` from my previous addition.  It was unnecessarily complicated and was a solution in search of a problem.  
 * Single valued ``Variables`` are graphed in ``Scalars`` else they are logged to ``Histogram`` and ``Distributions`` in TensorBoard. 
 * The ``tensorboard.ipynb`` has been updated and includes references to the gradient tracking and loss logged that was added in #598 

Thank you for all of your suggestions.  I believe this is much cleaner now.  I'm happy to modify anything else you'd like

Here's the new screenshots!
tensorboard_scalars

tensorboard_histograms

dustinvtran:
Great work. Merging now. I'll add a separate PR that adds the equivalent notebook as a tutorial on the website.

Issue #650, opened 2017-05-27
More Criticism (comparison to lme4) for Mixed Effects Model


patrickeganfoley:
Hello!

I just compared the mixed effects model example (https://github.com/blei-lab/edward/blob/master/notebooks/linear_mixed_effects_models.ipynb) to the same model in `lme4`, and was please to see they agree!  (I had checked b/c a very similar model looked like it wasn't converging.)

The instructor random effects, after 100 iterations, compared to those found by `lme4`:

![instructor_comparison_100](https://cloud.githubusercontent.com/assets/5103389/26518260/c693dab6-4261-11e7-8f1a-d32f972a07ce.png)

And the same comparison, after 10,000 iterations:

![instructor_comparison_10k](https://cloud.githubusercontent.com/assets/5103389/26518261/d2a3d284-4261-11e7-8919-a9885f7cf668.png)

Would it be cool if I added this comparison to the tutorial?

dustinvtran:
Yes! :)

patrickeganfoley:
Great!  I could either upload a cvs of the random effects fit by lme4, or call rpy2 within the notebook.  

Cool if I put the fit effects in /examples/data, or is there a better place?

dustinvtran:
That's fine for now. I think we'll eventually have an individual data subdirectory in `notebooks/`.

patrickeganfoley:
Cool!  Just opened a pr - https://github.com/blei-lab/edward/pull/655

dustinvtran:
Fixed in #655.

PR #655, opened 2017-05-28
Added comparison of lme4 fit to the linear_mixed_effects_model tutorial.

patrickeganfoley:
Per [this issue](https://github.com/blei-lab/edward/issues/650)

I added `lme4`'s estimates of the student, instructor, and department effects to `/examples/data`, and then added a few plots in the criticism section of the `linear_mixed_effects_models` tutorial comparing edward's fits to `lme4`s.  

I think they mostly match up OK (student and instructor effects very well) but the dept effects are a bit dissimilar.  I think it's ok to keep them in, but if it's too confusing I can also just drop them from both models, since the point is to show a mixed effects model in edward, not find the best fit for instructor ratings in the `InstEval` dataset.


dustinvtran:
Your conclusions about the departmental effects make sense. Happy to merge this in now. I'll add a separate commit that updates the corresponding `tex` file for the website.

patrickeganfoley:
Great!  Thanks!

PR #666, opened 2017-06-03
Fixed Exponential parameter in non-scalar base DP example

jengelman:
Noticed that the dirichlet process example doesn't finish, so updated the parameter names.


dustinvtran:
LGTN

Issue #635, opened 2017-05-04
variational inference: global step not exposed to custom optimizers


mariru:
By default, the variational inference class uses ADAM with decaying rate as optimizer.
In large models with many parameters using ADAM might be too costly, and using a basic gradient descent optimizer is more appropriate.
However, other optimizers can't use a decaying learning rate because `global_step` is not exposed.

A slight API change would be nice. 
One option is a flag that can be passed to  `inference.initialize`, like 'default_optimizer_adaptive' which by default is False. When it's True, the optimizer would default to the tf.GradientDescentOptimizer with decaying learning rate.

Alternatively one could add the option to pass `optimizer='gradientdescent_with_decay'` to `inference.initialize()`. 

Thoughts?

dustinvtran:
That makes sense. To keep the naming in line with TensorFlow, we could just call the argument `global_step` and which can be True or False.

kashif:
@dustinvtran there is already a var `global_step` in the function. Do you propose to rename it?

dustinvtran:
Good point. It would be bad if the argument is a bool and yet the internal TensorFlow function uses a `tf.Variable`. To keep things most general, I guess we could make the argument `global_step=None`, and the user can pass in a `tf.Variable`. This way `global_step` can be used across multiple inferences if desired(?). Other suggestions are welcome.

kashif:
cool... will check

kashif:
can be closed

PR #673, opened 2017-06-07
allow global_step to be passed as argument

kashif:
now all optimizers can use a decaying learning rate

for issue #635

### DIFF
--- a/edward/inferences/variational_inference.py
+++ b/edward/inferences/variational_inference.py
@@ -101,7 +101,7 @@ class VariationalInference(Inference):
     else:
       learning_rate = 0.01
       global_step = None
-      
+
     # Build optimizer.
     if optimizer is None:
       optimizer = tf.train.AdamOptimizer(learning_rate)
--- a/edward/inferences/variational_inference.py
+++ b/edward/inferences/variational_inference.py
@@ -32,7 +32,7 @@ class VariationalInference(Inference):
     super(VariationalInference, self).__init__(*args, **kwargs)
 
   def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
-                 *args, **kwargs):
+                 global_step=None, *args, **kwargs):
     """Initialize variational inference.
 
     Parameters
@@ -50,6 +50,8 @@ class VariationalInference(Inference):
       ``True`` if aim to use PrettyTensor optimizer (when using
       PrettyTensor) or ``False`` if aim to use TensorFlow optimizer.
       Defaults to TensorFlow.
+    global_step : a tf.Variable, optional
+      A TensorFlow variable to hold the global step.
     """
     super(VariationalInference, self).initialize(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


kashif:
Dear @dustinvtran kindly have a look at this PR, and do squash it if you merge eventually.

So the current change means that in all the examples the default optimizer is Adam without the decaying learning rate. So to get the old behaviour I will go through the sample code and add a  `global_step` variable and pass it along.

What do you say?

edward/inferences/variational_inference.py
```diff
@@ -87,38 +89,38 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
 
       self.summarize = tf.summary.merge_all(key=summary_key)
 
+    if isinstance(global_step, tf.Variable):
+        starter_learning_rate = 0.1
```
dustinvtran: We use two-space indent

edward/inferences/variational_inference.py
```diff
@@ -87,38 +89,38 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
...
+        optimizer = tf.train.FtrlOptimizer(learning_rate)
       elif optimizer == 'rmsprop':
-        optimizer = tf.train.RMSPropOptimizer(0.01)
+        optimizer = tf.train.RMSPropOptimizer(learning_rate)
       else:
         raise ValueError('Optimizer class not found:', optimizer)
-
-      global_step = None
     elif isinstance(optimizer, tf.train.Optimizer):
       # Custom optimizers have no control over global_step.
```
dustinvtran: This part in the logic can be removed.

kashif:
dear @dustinvtran happy for another review

edward/inferences/variational_inference.py
```diff
@@ -87,38 +89,36 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
 
       self.summarize = tf.summary.merge_all(key=summary_key)
 
-    if optimizer is None:
+    learning_rate = 0.01
```
dustinvtran: This line should be moved to L103 after checking `isinstance(optimizer, str)`.

edward/inferences/variational_inference.py
```diff
@@ -87,38 +89,36 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
...
+    if optimizer is None or isinstance(global_step, tf.Variable):
       # Use ADAM with a decaying scale factor.
-      global_step = tf.Variable(0, trainable=False, name="global_step")
+      if not global_step:
```
dustinvtran: This logic could be flatter/less nested.

I'm also not sure if it works. For example, suppose `isinstance(global_step, tf.Variable)` is True and the user passes in `optimizer` as a `str`. Then the logic will create the decaying learning rate and exit the if-else clause without building any optimizer.

dustinvtran: Suggestion: collapse the logic to be
```python
    if optimizer is None and global_step is None:
      # Default optimizer always uses a global step variable.
      global_step = tf.Variable(0, trainable=False, name="global_step")
    
    # Build learning rate step-size.
    if isinstance(global_step, tf.Variable):
      starter_learning_rate = 0.1
      learning_rate = tf.train.exponential_decay(starter_learning_rate,
                                                 global_step,
                                                 100, 0.9, staircase=True)
    else:
      learning_rate = 0.01

    # Build optimizer.
    if optimizer is None:
      optimizer = tf.train.AdamOptimizer(learning_rate)
    elif isinstance(optimizer, str):
       ...
```

kashif:
cool will do! My computer is with tech support... I hope I can pick it up
today and then I can push it.

edward/inferences/variational_inference.py
kashif: cool 💻  back from the 💀 ... anyways, so 1 question, do I remove the option that the  `optimizer` is of type `tf.train.Optimizer`, e.g. from the `TypeError` string and function doc?

Since it can now be either `None` or a string.

Thanks!

dustinvtran:
> do I remove the option that the optimizer is of type tf.train.Optimizer, e.g. from the TypeError string and function doc?
> 
> Since it can now be either None or a string.


That's correct. It should be converted back to `raise TypeError("Optimizer must be str or tf.train.Optimizer.")`

edward/inferences/variational_inference.py
```diff
@@ -50,6 +50,8 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
       ``True`` if aim to use PrettyTensor optimizer (when using
       PrettyTensor) or ``False`` if aim to use TensorFlow optimizer.
       Defaults to TensorFlow.
+    global_step : a tf.Variable, optional
```
dustinvtran: `global_step : tf.Variable, optional`

edward/inferences/variational_inference.py
```diff
@@ -87,38 +89,41 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
...
       self.summarize = tf.summary.merge_all(key=summary_key)
 
-    if optimizer is None:
-      # Use ADAM with a decaying scale factor.
+    if optimizer is None and global_step is None:
+      # Default optimizer always uses a global step variable.
+      global_step = tf.Variable(0, trainable=False, name="global_step")
+
+    if isinstance(global_step, tf.Variable):
       global_step = tf.Variable(0, trainable=False, name="global_step")
```
dustinvtran: Caught another typo. Looks like this line should be removed

edward/inferences/variational_inference.py
```diff
@@ -87,38 +89,41 @@ def initialize(self, optimizer=None, var_list=None, use_prettytensor=False,
...
       starter_learning_rate = 0.1
       learning_rate = tf.train.exponential_decay(starter_learning_rate,
                                                  global_step,
                                                  100, 0.9, staircase=True)
+    else:
+      learning_rate = 0.01
```
dustinvtran: After this line, we still have to set `global_step = None`.

kashif:
thank you @dustinvtran 🙇

PR #674, opened 2017-06-07
Feature/mc reset

matthewdhoffman:
This is a small PR that adds reset ops to the `MonteCarlo` class. This makes it possible to run a chain more than once.

My motivating use case is composing MCMC inference with other algorithms, which requires the ability to either run a chain indefinitely (e.g. using a circular buffer of samples) or the ability to run a fresh chain to convergence each iteration. I've got an ICML paper coming out soon that takes the latter approach, which I'm trying to reimplement in Edward—there'll be a few more PRs related to that coming.

### DIFF
--- a/tests/test-inferences/test_metropolishastings.py
+++ b/tests/test-inferences/test_metropolishastings.py
@@ -32,7 +32,6 @@ class test_metropolishastings_class(tf.test.TestCase):
       self.assertAllClose(qmu.stddev().eval(), np.sqrt(1 / 51),
                           rtol=1e-2, atol=1e-2)
 
-      sess = ed.get_session()
       old_t, old_n_accept = sess.run([inference.t, inference.n_accept])
       self.assertEquals(old_t, n_samples)
       self.assertGreater(old_n_accept, 0.1)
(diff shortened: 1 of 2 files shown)


tests/test-inferences/test_metropolishastings.py
```diff
@@ -31,6 +32,15 @@ def test_normalnormal_run(self):
       self.assertAllClose(qmu.stddev().eval(), np.sqrt(1 / 51),
                           rtol=1e-2, atol=1e-2)
 
+      sess = ed.get_session()
```
dustinvtran: You can remove this line since the `with` context already provides the session as `sess`.

dustinvtran:
That makes sense. I was also thinking of a similar feature for VI. For example, in (non-amortized) SVI, we need to reset `inference` when training a new set of local variables. I might add that in a PR after this one.

dustinvtran:
What are your thoughts on a reset op that users call, e.g., with `sess.run(inference.reset)` vs a slightly more abstract method call, e.g., with `inference.reset()`? This parallels `sess.run(inference.train)` vs `inference.update()`.

tests/test-inferences/test_metropolishastings.py
matthewdhoffman: Good catch. Fixed.

matthewdhoffman:
Good question. I guess it's probably good to have both, like we do with `inference.train` and `inference.update()`.

dustinvtran:
Merging now. 

Upon thinking more, not sure if we need the additional abstraction. Another parallel is how we rely on `tf.global_variables_initializer().run()` to initialize TensorFlow variables rather than write a new method. Access to reset ops is minimal and suits the job.

PR #678, opened 2017-06-12
WIP: multiple iters per sample

bfredl:
With long simulations, storing every Monte Carlo sample is not going to fit in memory, nor really meaningful. For instance with minibatches, for me storing samples per minibatch is too much, only once per full data should be enough.

This is the simple if somewhat hacky solution that uses `floor_div` to only increment the empirical index once every iters_per_sample. I guess another solution could be to allocate a separate variable for current state, and have a separate "sample" index that could be conditionally incremented or something (this could also be used to store nothing during burnin). This would make sense if we don't want to store the latent parameters at all. (Rather than storing the model, it might in some cases make sense to just store predictions of the model). What do you think would be the best?

### DIFF
--- a/edward/inferences/monte_carlo.py
+++ b/edward/inferences/monte_carlo.py
@@ -21,7 +21,8 @@ class MonteCarlo(Inference):
   minimum implement ``build_update``: it determines how to assign
   the samples in the ``Empirical`` approximations.
   """
-  def __init__(self, latent_vars=None, data=None):
+
+  def __init__(self, latent_vars=None, data=None, nsamples=1e4):
     """Initialization.
 
     Parameters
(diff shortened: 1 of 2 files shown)


dustinvtran:
I like your second suggestion. It's more extensible and thus has more applications. 

I think the only needed change is to include something like `sample_step` as an argument to `inference.initialize()`. It must be a scalar `tf.Tensor`, with a default of None, and internally will default to be equal to `self.t`. All MC algorithms when performing the assign op will write based on the sample index and not `self.t`.

An example use case is
```python
sample_step = tf.placeholder(tf.int32, [])
inference.initialize(sample_step=sample_step)
...
for _ in range(inference.n_iter):
  whatever_i_want = 12341234
  inference.update(feed_dict={sample_step: whatever_i_want})

PR #682, opened 2017-06-18
WIP  - Vanilla PCF of 20m MovieLens 

patrickeganfoley:
Hello!

Having some trouble with this and wanted to check to see if was doing anything very obviously wrong. 

I've set up a pretty vanilla (normal loss) PCF of the movie lens 20m data - 20 million ratings of 
of ~28k movies by ~135k users.  

Any tips on how to get this converging would be greatly appreciated!  Right now I'm thinking of trying

1.  use some batching
2.  switch to HPF following @dadaromeo 's excellent work here https://github.com/dadaromeo/recsys-hpf
3.  Wait?  try more learning rates?  

Any ideas?


patrickeganfoley:
(Link to notebook https://github.com/patrickeganfoley/edward/blob/master/notebooks/pcf_of_movielens_ratings.ipynb since the diff is not really viewable)

patrickeganfoley:
@dawenl Thanks for your help!

It's definitely converging now that I switched it from VI to MAP

loss_is_converging

However, it's not yet getting to results that I would call reasonable (e.g. the fit offsets seem to have no real correspondence to mean ratings in any way).  I'm going to just let it run for a few hours (have only been giving it 20m or so) before I take it another direction.

dawenl:
I am not sure if this is doable with edward, but traditionally when fitting these type of models, people usually just get a MLE point estimate for the offsets (i.e., no regularization). If you write out the MAP objective and derive the individual coordinate-wise updates, things might get a little more clear (and give you a idea of what to expect from edward).

Another crux with this type of bilinear model is that normally people either train it with closed-form coordinate updates or SGD by taking noisy gradient step by subsampling each (user, item) pair. What you are doing here is essentially full gradient descend. There is nothing wrong with it, but I wouldn't be surprised if you run into some convergence issue because of the sparsity (and power-law-ish property) of the data, which is why it is not a bad idea to start simple: MAP on user and item factors, and just plain MLE estimate for everything else. Maybe even do it in a coordinate-wise fashion.

patrickeganfoley:
@dawenl makes sense.  I tried running it coordinate-wise (separate inferences for user offsets, movie offsets, vecs, global regularization params, ...) and got the same results.  

I still think this ought to be doable in edward with regularization on all terms - and _hopefully_ without too much fine tuning.  I'm going to fool around w/ it a bit more (drop vecs so it's just IRT, make sure it's running correctly on a dense subset of the data, etc..) but will close this for now until I can get it working.  

Thanks for all your help!  

Any chance you're going to the BAPP meet up tomorrow?

dawenl:
I didn't know about this BAPP meetup. Looks like @dustinvtran is speaking! Interesting, I might stop by -- I live not far from Menlo Park.

PR #688, opened 2017-06-20
MXMX notebooks

xiangze:


xiangze:
I mistyped the title ... MXMX -> MCMC

dustinvtran:
Apologies for the delay. Is there a particular reason you think Bayesian linear/logistic regression with MCMC examples should be added? Since the change is a few lines from existing notebooks, maybe this would do better by making it a hierarchical model, and over a medium-size (~5,000 - 20,000 data points) real data set?

PR #692, opened 2017-06-24
Swap utf-8 progbar character for ASCII

rw:
In terminal work. printing Unicode characters can be problematic. In particular, using a Unicode character for the progressbar causes encoding errors if the environment is not set up properly:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\u2588' in position 15: ordinal not in range(128)
```

This patch changes the progress bar character to be ASCII, so that we have less to worry about when running Edward.

### DIFF
--- a/edward/util/progbar.py
+++ b/edward/util/progbar.py
@@ -77,7 +77,7 @@ class Progbar(object):
       bar += ' '
       prog_width = int(self.width * float(current) / self.target)
       if prog_width > 0:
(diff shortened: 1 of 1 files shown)


dustinvtran:
IMO the UTF-8 black bar is a lot cleaner. Does catching this error with a try/except block work? E.g., it will default to using the ASCII `*` if that error is raised.

PR #695, opened 2017-06-28
examples/factor_analysis.py: fixed outer for loop

akucukelbir:
The outer for loop had an off-by-one indexing bug.


dustinvtran:
LGTM

Issue #693, opened 2017-06-26
Errors in ICLR2017 demo


talbaumel:
Hi, When running the notebook on python 3.6 and latest versions of tensorflow, keras and edward I got the following error:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/miniconda3/lib/python3.6/site-packages/keras/engine/topology.py in assert_input_compatibility(self, inputs)
    418             try:
--> 419                 K.is_keras_tensor(x)
    420             except ValueError:

~/miniconda3/lib/python3.6/site-packages/keras/backend/tensorflow_backend.py in is_keras_tensor(x)
    391                           tf.SparseTensor)):
--> 392         raise ValueError('Unexpectedly found an instance of type `' + str(type(x)) + '`. '
    393                          'Expected a symbolic tensor instance.')

ValueError: Unexpectedly found an instance of type `<class 'abc.Normal'>`. Expected a symbolic tensor instance.

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
<ipython-input-2-2be4bc32d8a1> in <module>()
      8 # Probabilistic model
      9 z = Normal(loc=tf.zeros([N, d]), scale=tf.ones([N, d]))
---> 10 h = Dense(256, activation='relu')(z)
     11 x = Bernoulli(logits=Dense(28 * 28, activation=None)(h))
     12 

~/miniconda3/lib/python3.6/site-packages/keras/engine/topology.py in __call__(self, inputs, **kwargs)
    550                 # Raise exceptions in case the input is not compatible
    551                 # with the input_spec specified in the layer constructor.
--> 552                 self.assert_input_compatibility(inputs)
    553 
    554                 # Collect input shapes to build layer.

~/miniconda3/lib/python3.6/site-packages/keras/engine/topology.py in assert_input_compatibility(self, inputs)
    423                                  'Received type: ' +
    424                                  str(type(x)) + '. Full input: ' +
--> 425                                  str(inputs) + '. All inputs to the layer '
    426                                  'should be tensors.')
    427 

ValueError: Layer dense_1 was called with an input that isn't a symbolic tensor. Received type: <class 'abc.Normal'>. Full input: [<ed.RandomVariable 'Normal_2/' shape=(55000, 50) dtype=float32>]. All inputs to the layer should be tensors.
```

Thanks, Tal

dustinvtran:
[...]

PR #703, opened 2017-07-03
fixing example vae.py

romain-lopez:
Fixing example vae.py for keras compatibility 
https://github.com/blei-lab/edward/issues/693


dustinvtran:
LGTM

PR #705, opened 2017-07-03
Add example of LSTM language model for text8

dustinvtran:
Title. Two features are added relevant to the example:
+ A `maybe_download_and_extract` utility function. I've found it essential to automate loading of any standard data set. Here it's applied to text8 if the user doesn't have it.
+ `MAP` more intelligently avoids graph copying if `latent_vars` is empty. See also discussion on the Forum (https://discourse.edwardlib.org/t/limit-the-operations-that-are-duplicated-in-random-variables-py-copy/207).


lileicv:
I am Edward is a good library, Thank you.
I have a question. Can Edward deal with basicLSTMCell in tensorflow?
Hope to receive your reply.

PR #715, opened 2017-07-14
Implement a stochastic block model example

yamaguchiyuto:
This implements an example of stochastic block model on a toy data.


dustinvtran:
Cool! Since you're doing MAP jointly, is there a reason you split it up into an E and M step? It would also be really cool if you could try it on a standard data set, e.g., [Sampson's monastery data](https://networkdata.ics.uci.edu/netdata/html/sampson.html). As sanity check, you could compare to results like [the discussion around Fig 4 of the classic White et al. (1976)](http://www.stat.cmu.edu/~brian/780/bibliography/04%20Blockmodels/Wang%20-%201987%20-%20Stochastic%20blockmodels%20for%20directed%20graphs.pdf).

yamaguchiyuto:
Thanks for your response!

Changes are as follows:
* put the separated MAP together
* test the model using Zachary’s Karate Club graph
* follow PEP8

yamaguchiyuto:
I guess now this could be merged. How do you think?

dustinvtran:
Apologies for the delay. I took a look and would be happy to. One thing: can you move the data set into the `data/` subdirectory of `examples/`?

yamaguchiyuto:
Put them in `data/`. Could you take a look again?

dustinvtran:
+ Is `test.dat` used in the example script somewhere? 
+ Can you make sure that the code has 2-space indent?

yamaguchiyuto:
Thank you for additional comments!

- it was added by accident. deleted.
- fixed indent.

dustinvtran:
LGTM

RexYing:
Hi, when I try to run this example code, it seems that changing the seed from 42 to other numbers often makes the algorithm give bad results (rand index close to 0), even just for the karate club dataset, without changing anything else in the code. What might be the reason?

dustinvtran:
Have you tweaked hyperparameters? For example, 250 iterations seems a little arbitrary.

RexYing:
Increasing the # iterations up to 100k doesn't seem to help. I'm aware of label switching, but the problem seems to be that of mode collapsing.

For example, if I use the caveman graph with 4 communities as input, the output looks like:
[0 0 0 0 0 0 0 1 1 1 1 1 1 1 3 3 3 3 3 3 3 1 1 1 1 1 1 1]
The second community is collapsed with the first.

Sometimes when unlucky, multiple communities collapse.
Changing the initialization of parameters from uniform to uniformly random and doesn't help either.

joeporter:
@RexYing I know it has been several months for this thread, but I was seeing the exact same collapsing effect two days ago with the SBM algorithms in graph-tool (single cluster for some random seeds, full results with others).  I think it's an issue with the approach rather than the implementation.   If you've had any additional insights or observations, I would be interested.

rlouf:
@joeporter cf the issue on tensorflow-probability that is referenced above. Indeed, the approach is likely wrong and we need to correct the model for the heterogeneity of degrees in the graph. I am currently working on correcting this.

PR #717, opened 2017-07-19
Fixed weights' scale name in docs

stephenra:


dustinvtran:
Good catch. Can you also make the same change in `notebooks/tensorboard.ipynb`?

stephenra:
Done.

dustinvtran:
Thanks!

PR #726, opened 2017-08-01
Remove call to private method _set_device(), which may change.

matthewdhoffman:
I'm told that the behavior of this method may change in the not-too-distant future, so we shouldn't be calling it directly if it can be avoided.

I don't really understand what this code is doing, so I just deleted it. That doesn't seem to break any tests, but it probably isn't ideal. @dustinvtran, can you take a look?


dustinvtran:
This came from when we initially built `copy` via the function [`copy_op_to_graph`](https://github.com/tensorflow/tensorflow/blob/r1.3/tensorflow/contrib/copy_graph/python/util/copy_elements.py#L231) in `tf.contrib.copy_graph`. I'm not totally sure if this colocation device-setting is needed.

Issue #541, opened 2017-03-13
No U-Turn Sampler


dustinvtran:
Also see http://discuss.edwardlib.org/t/any-plans-for-nuts/.

korkinof:
I will eventually give it a try to implement this.
If anyone else wants to do this in the meantime, just give a shout here!

ssequeira:
I'd be happy to try my hand at this.

emilemathieu:
Hi !
I created a PR https://github.com/blei-lab/edward/pull/728 with the implementation of the Dual Averaging algorithm for adapting the step size of HMC, which is a first step to implement NUTS.

Razcle:
Hi I wondered if there was any update to this? I think there's probably a fair bit of demand for this feature and if help is required I'd like to chip in.

dustinvtran:
No need to put effort into. @axch has been working on this. Maybe he can reply on when something public-facing may appear in TF?

axch:
Don't know about time lines, exactly, but work is proceeding.

axch:
Actually, @dustinvtran, can you assign me this github issue (or give me sufficient permissions to assign it to myself)?

dustinvtran:
Gave you permission. Looks like you need to accept before assignments are available.

axch:
As an update, there is now a draft barebones NUTS implementation in TensorFlow Probability, at https://github.com/tensorflow/probability/tree/master/experimental/no_u_turn_sampler.  This implementation is expected to be compatible with Edward2.  See also the associated TFP issues https://github.com/tensorflow/probability/issues/125, https://github.com/tensorflow/probability/issues/126, https://github.com/tensorflow/probability/issues/142, and https://github.com/tensorflow/probability/issues/143.

axch:
TFP now has a batch NUTS (path length adaptation, haven't tested step size adaptation yet but don't anticipate excessive problems) here: https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/experimental/mcmc/nuts.py.

junpenglao:
NUTS sampler with multinominal sampling is currently on master: https://github.com/tensorflow/probability/commit/e57eba6c0ff15b2c94e0d802115641683206c38a

You can have a look at this end2end test of how to do sampling with dualaveraging step size adaptation: https://github.com/tensorflow/probability/blob/e57eba6c0ff15b2c94e0d802115641683206c38a/tensorflow_probability/python/mcmc/nuts_test.py#L284-L381

PR #728, opened 2017-08-03
Hamiltonian Monte Carlo with Dual Averaging

emilemathieu:
Hello to all !

This PR partially solve this issue https://github.com/blei-lab/edward/issues/541 which is about implementing [NUTS](https://arxiv.org/abs/1111.4246). As proposed in [this topic](https://discourse.edwardlib.org/t/any-plans-for-nuts/23), starting with the Dual Averaging is a good start.

The implemented inference method is in `edward/inferences/hmcda.py` and the corresponding test in `tests/test-inferences/test_hmcda.py`.

I hope you'll appreciate the PR ! :)

### DIFF
--- a/edward/inferences/hmcda.py
+++ b/edward/inferences/hmcda.py
@@ -1,0 +1,314 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import six
+import tensorflow as tf
+
+from collections import OrderedDict
+from edward.inferences.monte_carlo import MonteCarlo
+from edward.models import RandomVariable
+from edward.util import copy, get_session
+
+try:
+  from edward.models import Normal, Uniform
+except Exception as e:
+  raise ImportError("{0}. Your TensorFlow version is not supported.".format(e))
+
+
+class HMCDA(MonteCarlo):
+  """Hamiltonian Monte Carlo with Dual Averaging
+  (Hoffman M. and Gelman A., 2014) - Algorithm 5
+
+  Notes
+  -----
+  In conditional inference, we infer :math:`z` in :math:`p(z, \\beta
+  \mid x)` while fixing inference over :math:`\\beta` using another
+  distribution :math:`q(\\beta)`.
+  ``HMC`` substitutes the model's log marginal density
+
+  .. math::
+
+    \log p(x, z) = \log \mathbb{E}_{q(\\beta)} [ p(x, z, \\beta) ]
+                \\approx \log p(x, z, \\beta^*)
+
+  leveraging a single Monte Carlo sample, where :math:`\\beta^* \sim
+  q(\\beta)`. This is unbiased (and therefore asymptotically exact as a
+  pseudo-marginal method) if :math:`q(\\beta) = p(\\beta \mid x)`.
+  """
+  def __init__(self, *args, **kwargs):
+    """
+    Examples
+    --------
+    >>> z = Normal(loc=0.0, scale=1.0)
+    >>> x = Normal(loc=tf.ones(10) * z, scale=1.0)
+    >>>
+    >>> qz = Empirical(tf.Variable(tf.zeros(500)))
+    >>> data = {x: np.array([0.0] * 10, dtype=np.float32)}
+    >>> inference = ed.HMCDA({z: qz}, data)
+    """
+    super(HMCDA, self).__init__(*args, **kwargs)
+
+  def initialize(self, n_adapt, delta=0.65, Lambda=0.15, *args, **kwargs):
+    """
+    Parameters
+    ----------
+    n_adapt : float
+      Number of samples with adaption for epsilon
+    delta : float, optional
+      Target accept rate
+    Lambda : float, optional
+      Target leapfrog length
+    """
+    self.scope_iter = 0  # a convenient counter for log joint calculations
+  
+    # Find intial ϵ
+    step_size = self.find_good_eps()
+    sess = get_session()
+    init_op = tf.global_variables_initializer()
+    sess.run(init_op)
+    step_size = sess.run(step_size)
+
+    # Variables for Dual Averaging
+    self.epsilon = tf.Variable(step_size, trainable=False)
+    self.mu = tf.cast(tf.log(10.0*step_size), tf.float32)
+    self.epsilon_B = tf.Variable(1.0, trainable=False, name="epsilon_bar")
+    self.H_B = tf.Variable(0.0, trainable=False, name="H_bar")
+    
+    # Parameters for Dual Averaging
+    self.n_adapt = n_adapt
+    self.delta = tf.constant(delta, dtype=tf.float32)
+    self.Lambda = tf.constant(Lambda)
+    
+    self.gamma = tf.constant(0.05)
+    self.t_0 = tf.constant(10)
+    self.kappa = tf.constant(0.75)
+    
+    return super(HMCDA, self).initialize(*args, **kwargs)
+
+  def build_update(self):
+    """Simulate Hamiltonian dynamics using a numerical integrator.
+    Correct for the integrator's discretization error using an
+    acceptance ratio. The initial value of espilon is heuristically chosen
+    with Algorithm 4
+
+    Notes
+    -----
+    The updates assume each Empirical random variable is directly
+    parameterized by ``tf.Variable``s.
+    """
+    
+    old_sample = {z: tf.gather(qz.params, tf.maximum(self.t - 1, 0))
+                  for z, qz in six.iteritems(self.latent_vars)}
+    old_sample = OrderedDict(old_sample)
+
+    # Sample momentum.
+    old_r_sample = OrderedDict()
+    for z, qz in six.iteritems(self.latent_vars):
+      event_shape = qz.event_shape
+      normal = Normal(loc=tf.zeros(event_shape), scale=tf.ones(event_shape))
+      old_r_sample[z] = normal.sample()
+
+    # Simulate Hamiltonian dynamics.
+    L_m = tf.maximum(1, tf.cast(tf.round(self.Lambda / self.epsilon), tf.int32))
+    new_sample, new_r_sample = leapfrog(old_sample, old_r_sample,
+                                        self.epsilon, self._log_joint, L_m)
+
+    # Calculate acceptance ratio.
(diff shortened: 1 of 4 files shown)


emilemathieu:
@dustinvtran all checks have passed, could this PR be merged ?

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    Lambda : float, optional
+      Target leapfrog length
+    """
+    self.scope_iter = 0  # a convenient counter for log joint calculations
+
+    # Find intial epsilon
+    step_size = self.find_good_eps()
+    sess = get_session()
+    init_op = tf.global_variables_initializer()
+    sess.run(init_op)
```
dustinvtran: The initialize op shouldn't be needed inside `inference.initialize()`. It's called within the `run()` method or alternatively must be called manually after you call `inference.initialize()` on the algorithm.

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    Correct for the integrator's discretization error using an
+    acceptance ratio. The initial value of espilon is heuristically chosen
+    with Algorithm 4
+
+    Notes
+    -----
+    The updates assume each Empirical random variable is directly
+    parameterized by ``tf.Variable``s.
+    """
+
```
dustinvtran: remove emptyline

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    self.t_0 = tf.constant(10)
+    self.kappa = tf.constant(0.75)
+
+    return super(HMCDA, self).initialize(*args, **kwargs)
+
+  def build_update(self):
+    """Simulate Hamiltonian dynamics using a numerical integrator.
+    Correct for the integrator's discretization error using an
+    acceptance ratio. The initial value of espilon is heuristically chosen
+    with Algorithm 4
```
dustinvtran: `epsilon`, `Algorithm 4.`

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    n_adapt : float
+      Number of samples with adaption for epsilon
+    delta : float, optional
+      Target accept rate
+    Lambda : float, optional
+      Target leapfrog length
+    """
+    self.scope_iter = 0  # a convenient counter for log joint calculations
+
+    # Find intial epsilon
```
dustinvtran: `initial`

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    >>> inference = ed.HMCDA({z: qz}, data)
+    """
+    super(HMCDA, self).__init__(*args, **kwargs)
+
+  def initialize(self, n_adapt, delta=0.65, Lambda=0.15, *args, **kwargs):
+    """
+    Parameters
+    ----------
+    n_adapt : float
+      Number of samples with adaption for epsilon
```
dustinvtran: `adaptation`

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    # Calculate acceptance ratio.
+    ratio = kinetic_energy(old_r_sample)
+    ratio -= kinetic_energy(new_r_sample)
+    ratio += self._log_joint(new_sample)
+    ratio -= self._log_joint(old_sample)
+
+    # Accept or reject sample.
+    u = Uniform().sample()
+    alpha = tf.minimum(1.0, tf.exp(ratio))
+    accept = u < alpha
```
dustinvtran: Comparing with `tf.log(u) < ratio` should be more numerically stable than checking on the PDF scale.

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    # Update Empirical random variables.
+    for z, qz in six.iteritems(self.latent_vars):
+      variable = qz.get_variables()[0]
+      assign_ops.append(tf.scatter_update(variable, self.t, sample[z]))
+
+    # Increment n_accept (if accepted).
+    assign_ops.append(self.n_accept.assign_add(tf.where(accept, 1, 0)))
+    return tf.group(*assign_ops)
+
+  def do_not_adapt_step_size(self, alpha):
```
dustinvtran: for methods built for internal implementation, prepend the name with `_`.

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,331 @@
...
+    # Do not adapt step size but assign last running averaged epsilon to epsilon
+    assign_ops = []
+    assign_ops.append(self.H_B.assign_add(0.0).op)
```
dustinvtran: Is there a reason you add 0 to these variables? What happens if we don't use assign ops for them?

emilemathieu: The `tf.cond` arguments; `true_fn ` and `false_fn`must have the same type of outputs, which is a list of ops in our case.
We could also do ` assign_ops.append(tf.assign(self.H_B, self.H_B).op)`.
I would be happy to receive any better idea.

edward/inferences/hmcda.py
emilemathieu: Without this, I have: `Attempting to use uninitialized value Variable`.
I can't making it work without :/

emilemathieu:
Thanks for your feedbacks @dustinvtran !

I've fixed all your suggestions but the initialization issue. I dot know how to fix that.

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,330 @@
...
+    """Utility function to calculate model's log joint density,
+    log p(x, z), for inputs z (and fixed data x).
+
+    Parameters
+    ----------
+    z_sample : dict
+      Latent variable keys to samples.
+    """
+    self.scope_iter += 1
+    scope = 'inference_' + str(id(self)) + '/' + str(self.scope_iter)
```
dustinvtran: Note we no longer use `scope_iter` and `str(id(self))` implementation for scopes. See the latest versions of `hmc.py` and `sghmc.py` where we use `scope = tf.get_default_graph().unique_name("inference")`.

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,330 @@
...
+    # Update Empirical random variables.
+    for z, qz in six.iteritems(self.latent_vars):
+      variable = qz.get_variables()[0]
+      assign_ops.append(tf.scatter_update(variable, self.t, sample[z]))
+
+    # Increment n_accept (if accepted).
+    assign_ops.append(self.n_accept.assign_add(tf.where(accept, 1, 0)))
+    return tf.group(*assign_ops)
+
+  def _do_not__adapt_step_size(self, alpha):
```
dustinvtran: `_do_not_adapt_step_size`

dustinvtran:
> Without this, I have: `Attempting to use uninitialized value Variable.`
> I can't making it work without :/

Do you know which `tf.Variable` causes this to break?  For example, is there a reason you wrote `epsilon = tf.Variable(1.0, trainable=False)` instead of `epsilon = tf.constant(1.0)`?

emilemathieu:
I have updated to `epsilon = tf.constant(1.0)` but it still break without the initialization. 

Could it come from the empirical variables `self.latent_vars` which are needed in `find_good_eps` ?

dustinvtran:
Upon further investigation, the issue is data-dependent initialization. The `tf.Variable` `epsilon` depends on `tf.Variable`s in the model and approximating families. This means that the variables have to go through separate session calls to the init ops so that the model / approximating families are initialized first. Related: https://github.com/tensorflow/tensorflow/issues/4920

Not sure how to fix this just yet.

tests/test-inferences/test_hmcda.py
```diff
@@ -0,0 +1,36 @@
...
+
+      mu = Normal(loc=0.0, scale=1.0)
+      x = Normal(loc=tf.ones(50) * mu, scale=1.0)
+
+      qmu = Empirical(params=tf.Variable(tf.ones(5000)))
+
+      # analytic solution: N(loc=0.0, scale=\sqrt{1/51}=0.140)
+      inference = ed.HMCDA({mu: qmu}, data={x: x_data})
+      inference.run(n_adapt=1000)
+      print(qmu.mean().eval())
```
dustinvtran: remove print statements in test

edward/inferences/hmcda.py
```diff
@@ -0,0 +1,330 @@
...
+  z_new = z_old.copy()
+  r_new = r_old.copy()
+  keys_z_new = list(six.iterkeys(z_old.copy()))
+  first_grad_log_joint = tf.gradients(log_joint(z_new),
+                                      list(six.itervalues(z_new)))
+
+  k = tf.constant(0)
+
+  def while_condition(k, v_z_new, v_r_new, grad_log_joint):
+     # Stop when k < n_steps
```
dustinvtran: always use two space indent, including inside internal functions

emilemathieu:
Prints and indents fixed.

Nice spotted for the initialization issue ! What about the workaround proposed by yaroslavvb (2sd comment) ?

dustinvtran:
I hesitate because it (1) requires users to use a custom initialization scheme; and (2) depends on a `tf.contrib` library which can be unstable in its API / internal implementation. That said, I think it's worth trying as it may be the only viable solution; we can tweak/relax it later. 

When replacing the set of variables inside `inference.run()` with the following, I wasn't able to get it to work. You're welcome to tweak it so it does work.
```python
    if variables is None:
      # Force variables to be initialized after any variables they depend on.
      from tensorflow.contrib import graph_editor as ge
      def make_safe_initializer(var):
        """Returns initializer op that only runs for uninitialized ops."""
        return tf.cond(tf.is_variable_initialized(var),
                       tf.no_op,
                       lambda: tf.assign(var, var.initial_value).op,
                       name="safe_init_" + var.op.name).op

      safe_initializers = {}
      for v in tf.global_variables():
        safe_initializers[v.op.name] = make_safe_initializer(v)

      g = tf.get_default_graph()
      for v in tf.global_variables():
        var_name = v.op.name
        var_cache = g.get_operation_by_name(var_name + "/read")
        ge.reroute.add_control_inputs(var_cache, [safe_initializers[var_name]])

      init = tf.group(*safe_initializers.values())
    else:
      init = tf.variables_initializer(variables)
```

PR #732, opened 2017-08-09
Fix lint issues in test_map.py

kyleabeauchamp:
It seems like the tests might benefit from being run through a linter.  I happened to fix one of them while testing something else.

### DIFF
--- a/tests/test-inferences/test_map.py
+++ b/tests/test-inferences/test_map.py
@@ -11,21 +11,22 @@ from edward.models import Normal, PointMass
 
 class test_map_class(tf.test.TestCase):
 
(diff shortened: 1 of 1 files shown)


dustinvtran:
The current code a simple PEP8-style checker during Travis builds, subject to certain rules in [`setup.cfg`](https://github.com/blei-lab/edward/blob/master/setup.cfg) such as 2-space indent. You're welcome to use a linter to manually scan through the code base. But I don't think the linter would suggest any changes?

kyleabeauchamp:
Sorry about that, my linter wasn't reading the cfg.. The only other (non-spacing) suggestions were minor so I'll just close.

Issue #664, opened 2017-06-01
Multinomial classification metrics


dustinvtran:
Existing classification metrics assume the true and predicted output are each a single class label. We'd like to extend them to multinomial predictions where for each data point, we are bucketing a fixed number of trials into different categories.

cavaunpeu:
Hey @dustinvtran. I'd like to take this one. A few initial questions:

1. `if isinstance(output_key, binary_discrete + categorical_discrete)`, why would we ever use `n_samples > 1` to [generate `probs` in evaluate.py](https://github.com/blei-lab/edward/blob/91f63e49025006d0f5673f86835443fec6f89fc1/edward/criticisms/evaluate.py#L121)?

For multi-label data:

2. How should we [expand the `output_key` into `y_pred`](https://github.com/blei-lab/edward/blob/91f63e49025006d0f5673f86835443fec6f89fc1/edward/criticisms/evaluate.py#L128)? For instance, assume `output_key = Multinomial(total_count=5.0, probs=tf.constant([0.48, 0.51, 0.01]))`; what should `y_pred` be? What if `output_key = Multinomial(total_count=5.0, probs=tf.constant([0.33, 0.33, 0.34]))`?

3. Seemingly, the vanilla way to compute `categorical_accuracy` is to evaluate equivalence between `y_true` and `y_pred` along their outermost axes for each row (then return the mean). This is [how scikit-learn solves this problem](https://github.com/scikit-learn/scikit-learn/blob/d9fdd8b0d1053cb47af8e3823b7a05279dd72054/sklearn/metrics/classification.py#L171). Additionally, do we want to compute binary classification metrics *for each class* along this axis, then aggregate? Should the same aggregation strategies used in sckit-learn — dictated by the [`average` parameter in `_average_binary_score`](https://github.com/scikit-learn/scikit-learn/blob/d9fdd8b0d1053cb47af8e3823b7a05279dd72054/sklearn/metrics/base.py#L23) — be used?

4. Does it make sense to allow users to compute:
    - The mean squared error (and mean absolute error, etc.) between `y_true` and `y_pred`?
    - The categorical cross-entropy between something like `y_true / y_true.sum()` and [`probs`](https://github.com/blei-lab/edward/blob/91f63e49025006d0f5673f86835443fec6f89fc1/edward/criticisms/evaluate.py#L122)?

dustinvtran:
[...]

PR #743, opened 2017-08-30
Multinomial classification metrics

cavaunpeu:
Addresses [#664](https://github.com/blei-lab/edward/issues/664). Feedback appreciated, as I complete the second half.

Still to implement:
- [x] `compute_multinomial_mode`
- [x] scikit-learn-style binary classification aggregation metrics


edward/criticisms/evaluate.py
```diff
@@ -155,6 +160,10 @@ def evaluate(metrics, data, n_samples=500, output_key=None):
       evaluations += [categorical_crossentropy(y_true, y_pred)]
     elif metric == 'sparse_categorical_crossentropy':
       evaluations += [sparse_categorical_crossentropy(y_true, y_pred)]
+    elif metric == 'kl_divergence':
+      y_true_ = y_true / total_count
```
dustinvtran: This works only if it went through the specific logic above, i.e., `output_key` is categorical, multinomial, or onehotcategorical. Can you extend this to raise an error if it doesn't follow these classes—or more ambitiously, to have it work for all random variables?

edward/util/random_variables.py
```diff
@@ -699,3 +699,20 @@ def get_variables(x, collection=None):
...
   return list(output)
+
+
+def generate_multinomial_prediction(x, total_count=1):
+    """Generate a single multinomial prediction from a random variable,
+    comprised of `total_count` draws.
+
+    Args:
+      x: RandomVariable
+      total_count: number of draws to sample from `x`; the `n` in a Multinomial distribution.
```
dustinvtran: Is `total_count` the parameter in `x`, or is it the number of draws to sample from `x`? We refer to the latter by `sample_shape`. For example:
```python
Multinomial(total_count=2.0, probs=[0.25, 0.25, 0.5], sample_shape=5)
## <ed.RandomVariable 'Multinomial/' shape=(5, 3) dtype=float32>
```
represents 5 draws from a multinomial distribution.

edward/criticisms/evaluate.py
cavaunpeu: An umbrella question-as-a-response, which will help guide my changes: 

Do we want users, for example, to be able to compute `categorical_crossentropy([0, 2, 3], [0, 1, 4])`?

edward/util/random_variables.py
cavaunpeu: `total_count` is the parameter in `x`. I'll make this more clear.

edward/criticisms/evaluate.py
dustinvtran: Ideally no, because cross entropy is a function of two inputs: labels and logits/probabilities. Do we know what it means if the second input are also labels?

(Actually, going through the code, the way we currently calculate `*_crossentropy` and `hinge` seems wrong for discrete variables. They should be calculated on the discrete variables' logits/probs parameter and not the discrete variable itself.)

cavaunpeu: Why need cross-entropy be computed with labels and logits/probabilities? Isn't this purely a point of "mitigate-the-numerical-precision-issues" implementation?

cavaunpeu: On my end, I'm thinking of the question on a higher level. Cross-entropy gives the (pedantic definition to follow — apologies...) "uncertainty inherent in using the probabilities in one distribution to predict the events of another." To my knowledge, I don't believe these probabilities need even be normalized/valid. As such, given two tensors with outermost-dimensions of equal size, I don't see any mechanistic problem in allowing users to compute this metric or others similar at their own risk.

To this effect, I could simply add the following between lines 117 and [118](https://github.com/blei-lab/edward/blob/35423f22feb1d2598decbfcef0c021260ed4882a/edward/criticisms/evaluate.py#L118):

```python
total_count = tf.cond(hasattr(output_key, 'total_count'), lambda: output_key.total_count, lambda: 1)
```

To the contrary, should you want to take a more opinionated stance, i.e. precluding certain metrics from being calculated given certain random-variables, I'd think to do this at the beginning of the function. A priori, we'd assign a list of valid metrics as a property on each `RandomVariable` class; in `evaluate`, we'd loop through `metrics` and check for conflicts.

dustinvtran: Got it. Unnormalized probabilities are okay. But the probabilities input need to have float type. That is, your example question tries to calculate cross entropy for two integer-valued label inputs.

cavaunpeu: Why must they be floats? Isn't this a point of implementation as well?

I'll humbly interpret your comment to mean: as long as users pass inputs satisfying the `evaluate` API, they're free to evaluate with whatever metrics they want. Adding the snippet above between 117 and 118.

dustinvtran: That interpretation is reasonable. The only constraint is semantics: the second input _has_ to be a probability and not a label. With the current data types, discrete random variables with finite support and any `total_count` parameter must be integers and their probabilities must be floats.

cavaunpeu: Got it. The floats imperative is Tensorflow-specific, yeah? This part of your comment confused me at first, but the more I read Tensorflow docs, the more I realize how much it prefers things to be floats.

cavaunpeu: 👀/thoughts on `compute_multinomial_mode`?

Additionally, edward-specific stylistic and comments-best-practices feedback appreciated!

Finally, I'm assuming I'll squash these commits together before a potential merge.

dustinvtran: Looks great so far!

No need to manually squash. Edward does a squash+merge when merging any PRs.

williamabrwolf: Super. I'll finish up with "scikit-learn-style binary classification aggregation metrics" in the coming days.

cavaunpeu:
Just need to figure out how to get `mean_?_error` metrics to play nicely with `tf.constant(scalar)` inputs.

williamabrwolf:
@dustinvtran 

This should be ready for some eyes.

edward/criticisms/evaluate.py
```diff
@@ -330,9 +380,10 @@ def mean_squared_error(y_true, y_pred):
     y_pred: tf.Tensor.
       Tensors of same shape and type.
   """
-  return tf.reduce_mean(tf.square(y_pred - y_true))
+  return tf.reduce_mean(tf.square(y_pred - y_true), axis=-1)
```
dustinvtran: We always want metrics to return a scalar. This doesn't happen if, say, we're doing matrix regression where the output is a matrix, and we predict using a matrix and calculate MSE with this function.

edward/util/metrics.py
```diff
@@ -0,0 +1,35 @@
...
+from __future__ import division
+from __future__ import print_function
+
+
+from functools import wraps
+
+import tensorflow as tf
+
+
+def with_binary_averaging(metric):
```
dustinvtran: Great idea to include this. Can you explain how the function works as a wrapper? That is, how does a user specify the average option in `ed.evaluate`?

edward/util/random_variables.py
```diff
@@ -699,3 +699,52 @@ def get_variables(x, collection=None):
...
 
   return list(output)
+
+
+def softmax(vec):
+  numerator = np.exp(vec)
+  return numerator / numerator.sum(axis=0)
+
+
+def compute_multinomial_mode(probs, total_count=1):
```
dustinvtran: Is there motivation for writing this function in NumPy? In `evaluate.py`, it seems like you could keep `probs` as a `tf.Tensor` and make this function return the multinomial mode as a `tf.Tensor`.

edward/criticisms/evaluate.py
```diff
@@ -291,6 +307,39 @@ def sparse_categorical_crossentropy(y_true, y_pred):
...
+  vector of probabilities for `y_true`.
+
+  Args:
+    y_true: tf.Tensor.
+      Tensor of real values (probabilities) where the values in each row
+      of the outermost dimension sum to 1.
+    y_pred: tf.Tensor.
+      Same as `y_true`, and with the same shape.
+  """
+  y_true = tf.cast(y_true + 1.0, tf.float32)
```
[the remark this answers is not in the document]
cavaunpeu: `y_true` can plausibly include 0'; this would be problematic in taking the log.

Should I use `+ 1e-6` instead? (I feel like I'm missing something.)

edward/criticisms/evaluate.py
cavaunpeu: In fact, given the decorator, it always will return a scalar (unless `average=None` is specified). This logic plays out on lines 23 and 30 of "edward/util/metrics.py".

edward/util/metrics.py
cavaunpeu: Ah. Still need to add this. I'll add a test for it as well.

Thinking `ed.evaluate(['mean_absolute_error', ('mean_squared_error', {'average': 'micro'})], {x: x_data}, n_samples=1)`, for example.

edward/util/random_variables.py
cavaunpeu: Oh man. I tried for 2 days to implement this in TensorFlow.

Both the `while` loop — and updating `tf.Variable`s within — and the `np.random.choice` bit were *very* icky/overly-complicated in TensorFlow. 

I then realized that the only parameter this accepts is a vector. I asked myself if this vector (a probability distribution) would ever be exceptionally large, to which I answered unlikely. Finally, and at least for a first pass, I resolved that this vector would generally not be a multi-dimensional Tensor.

So, I went with Numpy, which made everything exceedingly straightforward.

edward/criticisms/evaluate.py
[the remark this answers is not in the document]
dustinvtran: The KL divergence for two discrete distributions p to q is a weighted sum of log ratios. Whenever p has zero probability in a term, the corresponding summand is zero. Whenever q has zero probability in a term, p must as well or else the KL divergence is undefined.

Assuming this is `KL(true || predict)`, you might implement this via:
```python
def ...:
  y_true = tf.cast(y_true, tf.float32)
  y_pred = tf.cast(y_pred, tf.float32)
  summand = tf.where(tf.equal(y_true, 0.0), 0.0, y_true * tf.log(y_true / y_pred))
  return tf.reduce_sum(summand)
```

edward/criticisms/evaluate.py
dustinvtran: Perfect!

edward/util/metrics.py
dustinvtran: Oof. It sounds difficult to specify intuitively. I like your approach for now; let's leave it as the default average option.

edward/util/random_variables.py
dustinvtran: Ah, yes that makes a lot of sense. NumPy is okay.

tests/test-util/test_random_variables.py
```diff
@@ -0,0 +1,25 @@
...
+
+import numpy as np
+import tensorflow as tf
+
+from edward.util.random_variables import compute_multinomial_mode
+
+
+class test_compute_multinomial_mode(tf.test.TestCase):
+
+    def test_correct_mode_computed_with_uniform_probabilities(self):
```
dustinvtran: This needs 2 space indent.

edward/criticisms/evaluate.py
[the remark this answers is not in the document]
cavaunpeu: Thanks, Dustin.

Is "whenever q has zero probability in a term, p must as well or else the KL divergence is undefined" codified in the formula above? Need it be?

edward/util/metrics.py
cavaunpeu: Working on this one now.

edward/criticisms/evaluate.py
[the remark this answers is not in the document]
dustinvtran: I predict the function will return a `tf.Tensor` evaluating to `np.nan` so it might be "implemented"; you can also check this in tests.

edward/util/metrics.py
williamabrwolf: Added.

cavaunpeu: Added.

cavaunpeu:
Okey. All set I think.

edward/criticisms/evaluate.py
```diff
@@ -16,16 +17,16 @@
...
   """Evaluate fitted model using a set of metrics.
 
-  A metric, or scoring rule [@winkler1994evaluating], is a function of
-  observed data under the posterior predictive distribution. For
-  example in supervised metrics such as classification accuracy, the
-  observed data (true output) is compared to the posterior
-  predictive's mean (predicted output). In unsupervised metrics such
-  as log-likelihood, the probability of observing the data is
-  calculated under the posterior predictive's log-density.
+  A metric, or scoring rule (Winkler, 1994), is a function of observed
```
dustinvtran: [@winkler1994evaluating]

edward/util/random_variables.py
```diff
@@ -778,3 +778,54 @@ def transform(x, *args, **kwargs):
   new_x = TransformedDistribution(x, bij, *args, **kwargs)
   new_x.support = new_support
   return new_x
+
+
+def softmax(vec):
```
dustinvtran: By convention, any functions without a leading underscore in `util/` are exposed to the user. Since you only apply this function once it could make sense to just write it in-line.

edward/util/metrics.py
```diff
@@ -0,0 +1,47 @@
...
+  """
+  Inspired by scikit-learn's _average_binary_score function:
+  http://bit.ly/2yhcABp.
```
dustinvtran: Short links are bad in docstrings because they're not transparent about the url. Can you paste the full link instead?

edward/util/metrics.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
```
dustinvtran: Single blankline instead of double here

edward/criticisms/evaluate.py
```diff
@@ -115,26 +116,43 @@ def evaluate(metrics, data, n_samples=500, output_key=None):
...
+      probs = np.sum(probs, axis=0) / n_samples
       if isinstance(output_key, binary_discrete):
         # make random prediction whenever probs is exactly 0.5
         random = tf.random_uniform(shape=tf.shape(probs))
         y_pred = tf.round(tf.where(tf.equal(0.5, probs), random, probs))
       else:
-        y_pred = tf.argmax(probs, len(probs.shape) - 1)
+        if total_count > 1:
+          mode = compute_multinomial_mode(probs, total_count, seed)
+          if len(output_key.sample_shape):
```
dustinvtran: Can you explain the use case for this part or provide an example? I'm not sure I followed this logic.

edward/criticisms/evaluate.py
```diff
@@ -115,26 +116,43 @@ def evaluate(metrics, data, n_samples=500, output_key=None):
...
       y_pred = tf.cast(tf.add_n(y_pred), y_pred[0].dtype) / \
           tf.cast(n_samples, y_pred[0].dtype)
+    if len(y_true.shape) == 0:
+      y_true = tf.expand_dims(y_true, 0)
+      y_pred = tf.expand_dims(y_pred, 0)
 
   # Evaluate y_true (according to y_pred if supervised) for all metrics.
   evaluations = []
   for metric in metrics:
+    if isinstance(metric, tuple):
```
dustinvtran: I personally don't think this API is intuitive and would default to not having the option exposed until we finalize ideas. That said, I don't feel too strongly about it so feel free to leave it here. However, do update the docstrings to reflect how to use this, and perhaps include an example in the docstrings' section.

edward/criticisms/evaluate.py
```diff
@@ -16,16 +17,16 @@
   raise ImportError("{0}. Your TensorFlow version is not supported.".format(e))
 
 
-def evaluate(metrics, data, n_samples=500, output_key=None):
+def evaluate(metrics, data, n_samples=500, output_key=None, seed=None):
```
dustinvtran: Need to update docstring with this new arg. You can follow the standard arg description of seed from TF docstrings.

edward/util/random_variables.py
cavaunpeu: Makes sense. Moved this into `compute_multinomial_mode`. Left it as a function, as it's used in the `while` loop.

edward/criticisms/evaluate.py
cavaunpeu: Done.

edward/criticisms/evaluate.py
cavaunpeu: Done.

edward/criticisms/evaluate.py
dustinvtran: red text

dustinvtran:
LGTM. Thanks again!

PR #758, opened 2017-09-18
Feature/automated transformations

cshenton:
## Random Variable transform utility

Adds `ed.util.transform`, which automates transformation of arbitrary continuous RVs to an unbounded space. Also provides a convenience method for applying arbitrary bijections to input RVs. 

This function is also available as `ed.transform` for convenience.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,8 +100,3 @@ docs/*.html
 # IDE related
 .idea/
 .vscode/
-
-# virtualenv
-.venv
-venv/
-ENV/
--- a/.gitignore
+++ b/.gitignore
@@ -100,3 +100,8 @@ docs/*.html
 # IDE related
 .idea/
 .vscode/
+
+# virtualenv
+.venv
+venv/
+ENV/
--- a/edward/util/random_variables.py
+++ b/edward/util/random_variables.py
@@ -10,7 +10,7 @@ from copy import deepcopy
 from edward.models.random_variable import RandomVariable
 from edward.models.random_variables import TransformedDistribution
 from edward.util.graphs import random_variables
-from tensorflow.contrib.distributions import bijector
+from tensorflow.contrib.distributions import bijectors
 from tensorflow.core.framework import attr_value_pb2
 from tensorflow.python.framework.ops import set_shapes_for_outputs
 from tensorflow.python.util import compat
@@ -745,15 +745,24 @@ def transform(x, *args, **kwargs):
   """
   if len(args) != 0 or kwargs.get('bijector', None) is not None:
     return TransformedDistribution(x, *args, **kwargs)
-  elif x.support == '01':
-    bij = bijector.Invert(bijector.SigmoidCentered())
-  elif x.support == 'nonnegative':
-    bij = bijector.Invert(bijector.Softplus())
-  elif x.support == 'simplex':
-    bij = bijector.Invert(bijector.SoftmaxCentered())
-  elif x.support == 'real' or 'multivariate_real':
+
+  try:
+    support = x.support
+  except AttributeError as e:
+    msg = "'{}' object has no 'support' so cannot be transformed.".format(
+      type(x).__name__)
+    raise NotImplementedError(msg)
+
+  if support == '01':
+    bij = bijectors.Invert(bijectors.Sigmoid())
+  elif support == 'nonnegative':
+    bij = bijectors.Invert(bijectors.Softplus())
+  elif support == 'simplex':
+    bij = bijectors.Invert(bijectors.SoftmaxCentered(event_ndims=1))
+  elif support == 'real' or support == 'multivariate_real':
     return x
   else:
+    msg = "'transform' does not handle supports of type '{}'".format(support)
     raise NotImplementedError()
 
   return TransformedDistribution(x, bij, *args, **kwargs)
--- a/edward/util/random_variables.py
+++ b/edward/util/random_variables.py
@@ -743,10 +743,8 @@ def transform(x, *args, **kwargs):
   >>> sess.run(y)
   -2.2279539
   """
-  if len(args) != 0:
-    bij = args.pop(0)
-  elif kwargs.get('bijector', None) is not None:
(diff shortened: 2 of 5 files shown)


cshenton:
Regarding broader support for supports, I think the current approach of their being a string property may be limiting. Since, for example, to transform `Uniform(a, b)`, I want to use a,b to get `Uniform(0, 1)` and then apply a sigmoid transform.

Perhaps we could implement something like:
```python
>>> u = Uniform(2.0, 3.0)
>>> u.support.eval()
[2.0, 3.0]
>>> u.support_type
"bounded"
>>> g = Gamma(1.0, 1.0)
>>> g.support
[0.0, np.inf]
>>> g.support_type
"nonnegative"
```

tests/test-util/test_transform.py
```diff
@@ -0,0 +1,75 @@
...
+    with self.test_session():
+      x = Normal(0.0, 1.0)
+      y = ed.transform(x, bijectors.Softplus())
+      y.eval()
+
+  def test_kwargs(self):
+    with self.test_session():
+      x = Normal(0.0, 1.0)
+      y = ed.transform(x, bijector=bijectors.Softplus())
+      y.eval()
```
dustinvtran: Can you check that the value is non-negative? For example, you might start with a normal distribution with mean of -100.0, standard deviation 0.1. Then assert that 10 draws from the transformed distribution are all non-negative.

Same goes for the other draws.

edward/util/random_variables.py
```diff
@@ -712,3 +714,57 @@ def get_variables(x, collection=None):
...
 
   return list(output)
+
+
+def transform(x, *args, **kwargs):
+  """Transform a continuous random variable to be on the unconstrained
+  space. It selects according to a number of default transformations,
+  depending on the support.
+
+  Parameters
```
dustinvtran: Our docstring format changed to follow TensorFlow's. Can you make that change? (c.f., the other function docstrings in this file)

edward/util/random_variables.py
```diff
@@ -712,3 +714,57 @@ def get_variables(x, collection=None):
...
+  if len(args) != 0 or kwargs.get('bijector', None) is not None:
+    return TransformedDistribution(x, *args, **kwargs)
+
+  try:
+    support = x.support
+  except AttributeError as e:
+    msg = """'{}' object has no 'support'
+             so cannot be transformed.""".format(type(x).__name__)
+    raise ValueError(msg)
+    # TODO: Add supports to RandomVariables missing them
```
dustinvtran: Can you remove TODOs?

.gitignore
```diff
@@ -100,3 +100,8 @@ docs/*.html
 # IDE related
 .idea/
 .vscode/
+
```
dustinvtran: Can you remove these changes to `.gitignore`? It could make sense to add these but it's not related to this PR.

cshenton:
Thanks for the comments Dustin, I will implement those changes now.

Thanks also for linking that design doc, will have a more thorough read later but it seems very thoughtful so far. Where's the best place to put any thoughts I have? Also, would your preference be to implement fuller featured supports in Edward, then to PR upstream once the API settles down, or would it be to directly PR the feature to tensorflow?

.gitignore
cshenton: Will do. Will have a think about it and might make a separate PR if it still makes sense on second thought.

tests/test-util/test_transform.py
cshenton: Good idea, I'll do that.

More generally, do you have any opinion regarding testing of probabilistic components? I'll just do a single sample check for now, but more would be better. It might be worth having a rule of thumb guideline to apply across the codebase (say, 20 samples). Even better would be an importable global variable that can be set when calling the test runner.

cshenton:
Okay, for each automatic transform, I take ten draws with a fixed seed and verify that there is both a positive and negative draw in those samples. It's the closest approximation I could think of to test for an unbounded support.

For the multivariate distributions, I've taken the same approach, but I am wondering if it would be more correct to check for positive and negative elements along each dimension, rather than in the sampled array as a whole. Let me know if that's preferable.

akucukelbir:
Hi Charles. This comes at a great time. I finally have a bit of time to spend on getting back into Edward dev! 

> For the multivariate distributions, I've taken the same approach, but I am wondering if it would be more correct to check for positive and negative elements along each dimension, rather than in the sampled array as a whole. Let me know if that's preferable.

I think this would be a good idea. As dimensions increase, checking the sampled array as a whole may lead to bugs crawling through the cracks.

dustinvtran:
> Where's the best place to put any thoughts I have? Also, would your preference be to implement fuller featured supports in Edward, then to PR upstream once the API settles down, or would it be to directly PR the feature to tensorflow?

I suggest opening a Github issue on tensorflow. Others including me can add thoughts about the design. Importantly, we should finalize it before anyone makes specific code changes which could end up not being merged at all.

> Okay, for each automatic transform, I take ten draws with a fixed seed and verify that there is both a positive and negative draw in those samples. It's the closest approximation I could think of to test for an unbounded support.

That makes sense. re:multivariate discussion. Maybe what can do is instantiate a `sample_shape` of say, 10, to represent the 10 draws. Then check that there are positive and negative elements along each dimension in the array; this holds for both multivariate and univariate distributions.

cshenton:
@akucukelbir that's great to hear. Your input on implementing advi in Edward will be invaluable.
@dustinvtran okay, I will open an issue when I have a moment, and link back to it here for reference. Would you be alright if I reproduced your code samples (from the google doc) with credit as a way to start the discussion?

In the mean time, I'll make those fixes to multivariate draws so we can finish up this feature.

cshenton:
Okay, wrote a small method that reduces along the sample dimension, then asserts each element along each other dimension satisfies the criterion (i.e. contains one element < 0, one element > 0). So support checks on multidimensional distributions are more correct now.

dustinvtran:
LGTM. Looking forward to its use in VI and MC algorithms.

PR #761, opened 2017-09-22
Add automated transformations to all inference

dustinvtran:
todos
+ [x] How do we handle discrete support in inference?
  + (For now, we don't do anything as discrete variables don't have the `support` attribute.)
+ [x] How do we store transformed vs original vars? especially to obtain constrained approximations (softplus-normal) when user defines unconstrained approximations (normal)
  + We use an additional member `self.transformations`.
+ [x] How does the implementation work for pointmass when doing MAP and normal when doing Laplace?
  + MAP optimizes over constrained support with unconstrained free parameters. Normal does the same and calculates the Hessian over the constrained log-density on the constrained mode.
+ [x] How does the implementation work for empirical when doing monte carlo algs?
  + Monte Carlo algs restricted to real support like HMC run a Markov chain on the unconstrained space and write samples in Empirical's via their constrained space.
+ [ ] Automated transformations tutorial

### DIFF
--- a/edward/inferences/inference.py
+++ b/edward/inferences/inference.py
@@ -218,19 +218,10 @@ class Inference(object):
       latent_vars = self.latent_vars.copy()
       self.latent_vars = {}
       for z, qz in six.iteritems(latent_vars):
-        try:
-          if z.support != qz.support:
-            z_transform = transform(z)
-            if qz.support == 'point':
-              from tensorflow.contrib.distributions import bijectors
-              qz_transform = transform(qz, bijectors.Invert(z_transform.bijector))
-              self.latent_vars[z] = qz_transform
-            else:
-              qz_transform = transform(qz)
-              self.latent_vars[z_transform] = qz_transform
-          else:
-            self.latent_vars[z] = qz
-        except AttributeError:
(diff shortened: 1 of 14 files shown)


edward/inferences/inference.py
```diff
@@ -210,6 +214,18 @@ def initialize(self, n_iter=1000, n_print=None, scale=None, logdir=None,
 
     self.scale = scale
 
+    if auto_transform:
+      latent_vars = self.latent_vars.copy()
+      self.latent_vars = {}
+      for z, qz in six.iteritems(latent_vars):
+        try:
+          if z.support != qz.support:
```
akucukelbir: If the supports are matched, nothing happens.
If the supports are not matched, individual z,qz pairs get transformed.

Does it make sense to log/print/report these somehow? I can see this getting challenging to debug going forward?

dustinvtran: Yep, I agree. It's sort of open how to report these automations so we haven't gotten around to it. Other examples are the default optimizer and the default posterior approximations in ed.MonteCarlo and ed.MAP.

Maybe we can establish such a standard in a new PR. I added an issue #764.

PR #769, opened 2017-09-27
Renyi divergence

jbregli:
Here is an implementation of the Renyi divergence variational inference.
There's also an example on VAEs.

Here is a link to the edward forum with some more info:
https://discourse.edwardlib.org/t/renyi-divergence-variational-inference/366/3


.gitignore
```diff
@@ -100,3 +100,9 @@ docs/*.html
 # IDE related
 .idea/
 .vscode/
+
```
dustinvtran: Can you remove changes that aren't relevant for this PR? This includes changes to `.gitignore` here as well as deletion of CSVs.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,184 @@
...
+import six
+import numpy as np
+import tensorflow as tf
+
+from edward.inferences.variational_inference import VariationalInference
+from edward.models import RandomVariable
+from edward.util import copy
+
+try:
+    from edward.models import Normal
```
dustinvtran: As convention, we use 2-space indent.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,184 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import six
+import numpy as np
```
dustinvtran: As convention, we alphabetize the ordering of the import libraries.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,184 @@
...
+
+try:
+    from edward.models import Normal
+    from tensorflow.contrib.distributions import kl_divergence
+except Exception as e:
+    raise ImportError(
+        "{0}. Your TensorFlow version is not supported.".format(e))
+
+
+class Renyi_divergence(VariationalInference):
```
dustinvtran: As convention, we use CamelCase for class names.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,184 @@
...
+    """Variational inference with the Renyi divergence
+
+    $ \text{D}_{R}^{(\alpha)}(q(z)||p(z \mid x))
+        = \frac{1}{\alpha-1} \log \int q(z)^{\alpha} p(z \mid x)^{1-\alpha} dz $
+
+    To perform the optimization, this class uses the techniques from
+    Renyi Divergence Variational Inference (Y. Li & al, 2016)
+
+    # Notes:
```
dustinvtran: Docstrings are parsed as Markdown and formatted in a somewhat specific way as they appear on the API docs. I recommend following the other classes, where you would denote a subsection as `#### Notes` and when writing bullet points, do, e.g.,
```
#### Notes

+ bullet 1
+ bullet 2
  + maybe bulleted list in a bullet
```

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+        "{0}. Your TensorFlow version is not supported.".format(e))
+
+
+class Renyi_divergence(VariationalInference):
+    """Variational inference with the Renyi divergence
+
+    $ \text{D}_{R}^{(\alpha)}(q(z)||p(z \mid x))
+        = \frac{1}{\alpha-1} \log \int q(z)^{\alpha} p(z \mid x)^{1-\alpha} dz $
+
+    To perform the optimization, this class uses the techniques from
```
dustinvtran: Periods at end of sentences. (If you'd look at the generated API for the class, I recommend compiling the website following instructions from `docs/`.)

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+    Renyi Divergence Variational Inference (Y. Li & al, 2016)
```
dustinvtran: We use bibtex for handling references in docstrings. This is handled by adding the appropriate bib entry to `docs/tex/bib.bib`; make sure it's also written in the right order: we sort bib entries by their year, then alphabetically according to their citekey within each year.

When using references, you can produce `(Li et al., 2016)` and `Li et al. (2016)` by writing `[@li2016renyi]` and @li2016renyi` respectively, assuming that `li2016renyi` is the citekey.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # Notes:
+        - Renyi divergence does not have any analytic version.
+        - Renyi divergence does not have any version for non reparametrizable
```
dustinvtran: It does but the gradient estimator in `@li2016variational` doesn't.  I recommend just stating that this inference algorithm is restricted to variational approximations whose random variables all satisfy `rv.reparameterization_type == tf.contrib.distributions.FULLY_REPARAMETERIZED`.

Also, instead of checking this during `build_loss_and_gradients` I recommend checking this during the `__init__`. This sort of check is done statically any graph construction similar to how we check for compatible shapes in all latent variables and data during `__init__`.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+            anymore. This mode is not describe in the paper but implemented
+            in the publicly available implementation of the paper's experiments.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super(Renyi_divergence, self).__init__(*args, **kwargs)
+
+    def initialize(self,
+                   n_samples=32,
+                   alpha=1.,
```
dustinvtran: As convention, we append all numerics with 0, e.g., `1.0`.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+                   *args, **kwargs):
+        """Initialize inference algorithm. It initializes hyperparameters
+        and builds ops for the algorithm's computation graph.
+
+        Args:
+            n_samples: int, optional.
+                Number of samples from variational model for calculating
+                stochastic gradients.
+            alpha: float, optional.
+                Renyi divergence coefficient.
```
dustinvtran: Could be useful to specify the domain of the coefficient. E.g., `Must be greater than 0.` or etc.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+
+            grads = tf.gradients(loss, var_list)
+            grads_and_vars = list(zip(grads, var_list))
+            return loss, grads_and_vars
+        else:
+            raise NotImplementedError(
+                "Variational Renyi inference only works with reparameterizable"
+                " models")
+
+#########
```
dustinvtran: This function is only used in one location and is a one-liner; could you write that line instead of defining a new function?

examples/vae_renyi.py
```diff
@@ -0,0 +1,91 @@
...
+# INFERENCE
+# Define a subgraph of the variational model, corresponding to a
+# minibatch of size M.
+x_ph = tf.placeholder(tf.int32, [M, 28 * 28])
+hidden = Dense(256, activation='relu')(tf.cast(x_ph, tf.float32))
+qz = Normal(loc=Dense(d)(hidden),
+            scale=Dense(d, activation='softplus')(hidden))
+
+# Bind p(x, z) and q(z | x) to the same TensorFlow placeholder for x.
+inference = Renyi_divergence({z: qz}, data={x: x_ph})
```
dustinvtran: This code looks exactly the same as an older version of `vae.py` but only differs in this line. To keep the VAE versions better synced, could you add a comment suggesting that this is also an alternative in the existing `vae.py`?

Ideally, we'd like a specific application where `ed.RenyiDivergence` produces better results by some metric than alternatives. IIRC, the paper had some interesting results for a Bayesian neural net on some specific UCI data sets. That would be great to have and reproduce some of their results.

If you don't have time for this, we can leave it off for now and raise it as a Github issue post-merging this PR.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,191 @@
...
+                        * z_copy.log_prob(dict_swap[z]))
+
+                for x in six.iterkeys(self.data):
+                    if isinstance(x, RandomVariable):
+                        x_copy = copy(x, dict_swap, scope=scope)
+                        p_log_prob[s] += tf.reduce_sum(
+                            self.scale.get(x, 1.0)
+                            * x_copy.log_prob(dict_swap[x]))
+
+            logF = [p - q for p, q in zip(p_log_prob, q_log_prob)]
```
dustinvtran: Instead of `logF`, what about something like `log_ratios`, which is more Pythonic in snake_case and also more semantically meaningful?

jbregli:
Thanks for the suggestion and the very informative feedback.

jbregli:
I've added some testing in a similar way as KLqp. (both normal_normal and the bernouilli distribution.
For each cases, I've tested most of the possible cases of the Renyi VI:
KL, VR-max, VR-min, alpha<0, alpha>0.

tests/inferences/test_renyi_divergence.py
```diff
@@ -0,0 +1,121 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import edward as ed
+import numpy as np
+import tensorflow as tf
+
+from edward.models import Bernoulli, Normal
+from edward.inferences.renyi_divergence import RenyiDivergence
```
dustinvtran: We should check the import also works by using `ed.RenyiDivergence` in the test.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,176 @@
...
+  It minimizes the Renyi divergence
+
+  $ \text{D}_{R}^{(\alpha)}(q(z)||p(z \mid x))
+      = \frac{1}{\alpha-1} \log \int q(z)^{\alpha} p(z \mid x)^{1-\alpha} dz.$
+
+  The optimization is performed using the gradient estimator as defined in
+  [@li2016renyi].
+
+  #### Notes
+      + The gradient estimator used here does not have any analytic version.
```
dustinvtran: With Markdown formatting, you don't need the 4 spaces of indentation. E.g., you can just do
```
#### Notes

+ The gradient estimator ...
+ ...
```

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,176 @@
...
+class RenyiDivergence(VariationalInference):
+  """Variational inference with the Renyi divergence [@li2016renyi].
+
+  It minimizes the Renyi divergence
+
+  $ \text{D}_{R}^{(\alpha)}(q(z)||p(z \mid x))
+      = \frac{1}{\alpha-1} \log \int q(z)^{\alpha} p(z \mid x)^{1-\alpha} dz.$
+
+  The optimization is performed using the gradient estimator as defined in
+  [@li2016renyi].
```
dustinvtran: The citekey is being used as a direct object so it should be `[@li2016renyi]` -> `@li2016renyi`.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,176 @@
...
+        + Reparameterization gradients [@kingma2014auto].
+        + Stochastic approximation of the joint distribution [@li2016renyi].
+
+    #### Notes
+        + If the model is not reparameterizable, it returns a
+        NotImplementedError.
+        + See Renyi Divergence Variational Inference [@li2016renyi] for
+        more details.
+    """
+    if self.is_reparameterizable:
```
dustinvtran: `is_reparameterizable` should be checked with the possible raising error during the `__init__`, and since it's checked there it doesn't need to be stored in the class. This also helps to remove one layer of indentation in this function.

edward/inferences/renyi_divergence.py
```diff
@@ -0,0 +1,176 @@
...
+            x_copy = copy(x, dict_swap, scope=scope)
+            p_log_prob[s] += tf.reduce_sum(
+                self.scale.get(x, 1.0) * x_copy.log_prob(dict_swap[x]))
+
+      log_ratios = [p - q for p, q in zip(p_log_prob, q_log_prob)]
+
+      if self.backward_pass == 'max':
+        log_ratios = tf.stack(log_ratios)
+        log_ratios = tf.reduce_max(log_ratios, 0)
+        loss = tf.reduce_mean(log_ratios)
```
dustinvtran: If I understood the code correctly, `log_ratios` when first created is a list of `n_samples` elements, where each element is a log ratio calculation per sample from q. For the min / max modes, we take the min / max of these log ratios, which is a scalar.

Is `tf.reduce_mean` for the loss needed? You can also remove the `tf.stack` line in the min and max cases in the same way you didn't use it for the `self.alpha \approx 1` case.

examples/vae_renyi.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
+"""Variational auto-encoder for MNIST data using Renyi variational objective
+[@li2016renyi]
+
+#### Notes
+This example is almost exactly similar to example/vae.py.
```
dustinvtran: Sorry for the miscommunication. What I meant was that you can edit `vae.py`, comment out the 1-2 lines of code to use `ed.RenyiDivergence`, and add these notes there. This helps to compress the content in the examples, c.f., https://github.com/blei-lab/edward/blob/master/examples/bayesian_logistic_regression.py#L51.

tests/inferences/test_renyi_divergence.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class test_renyi_divergence_class(tf.test.TestCase):
+
+  def _test_normal_normal(self, Inference, *args, **kwargs):
```
dustinvtran: Since `RenyiDivergence` is used across all tests, you don't need it as an arg to the test functions.

tests/inferences/test_renyi_divergence.py
jbregli: Got some issue with this but it should be working now.

tests/inferences/test_renyi_divergence.py
jbregli: I've used the same template as `test_klpq` where only `KLpq` is used during the tests and where `Inference` is stil an argument to the test functions.
But I think I have modified to be closer to what you had in mind

examples/vae_renyi.py
jbregli: Removed vae_renyi.py and modifed vae.py instead.
The version of vae.py I had wasn't running though. So I've modified it quite a bit.

edward/inferences/renyi_divergence.py
jbregli: You're right. Thanks for spotting this.

edward/inferences/renyi_divergence.py
jbregli: Done

edward/inferences/renyi_divergence.py
jbregli: Done

edward/inferences/renyi_divergence.py
jbregli: Done

jbregli:
It keep failing the travis-ci check for python 2.7 but before getting into the proper testing of my code (fail to install matplotlib and seaborn). 
Anything I've done wrong on my side?

examples/vae_renyi.py
dustinvtran: Are you using the latest version of Edward? We updated a few details in vae.py so it actually runs better. For example, you should be using the observations library and a generator, which is far more transparent than the `mnist_data` class from that TensorFlow tutorial.

In addition, since `vae.py` is also our canonical VAE example, I prefer keeping it as `ed.KLqp` as the default, and with the renyi divergence option commented out; similarly, the top-level comments should be written in-line near the renyi divergence option instead.

If you have thoughts otherwise, happy to take alternative suggestions.

dustinvtran:
Looks like this is happening in Travis on any build. I'll look into it.

Issue #759, opened 2017-09-22
Efficient Full Model Sampling


cshenton:
`Distribution.sample()` evaluates all distribution parameters, and then samples from the resulting distribution, this means that if parameters are RVs, only one sample is taken. For 'full model' sampling (letting the observed and unobserved RVs vary freely) the current approach seems to be [repeatedly passing through the graph](https://github.com/blei-lab/edward/blob/master/edward/criticisms/ppc.py#L112).

However `Distribution` permits parameters to be passed through as a batch, which could produce full model variation if desired. The idea would be to let the ser define the model in the most natural way then (for example) for RandomVariable to overload `sample()` to recursively sample down the graph. This could potentially make inference methods that rely on full model sampling more efficient too. 

Let me know if this is currently possible and I've missed something, but I found myself explicitly writing sample methods for some of my models (to do just this) and wondered if it couldn't just happen automatically.

For example
```python
import edward as ed
import matplotlib.pyplot as plt
import tensorflow as tf

from edward.models import Normal

sess = ed.get_session()

# Naive approach (single mu draw)
mu = Normal(0.0, 5.0)
sd = 0.001
y = Normal(mu, sd)
sample = y.sample(1000).eval()

# Current approach (repeated graph evaluation)
mu = Normal(0.0, 5.0)
sd = 0.001
y = Normal(mu, sd)
sample = [y.eval() for _ in range(1000)]

# Behaviour we want to emulate (many mu draws, probably much faster)
mu = Normal(0.0, 5.0, sample_shape=[1000])
sd = 0.001
y = Normal(mu, sd)
sample = y.sample().eval()

# Possible API for that behaviour
mu = Normal(0.0, 5.0)
sd = 0.001
y = Normal(mu, sd)
y.sample(1000, deep=True).eval()
```
At eval, instead of evaluating mu, it samples from mu then passes in a batch parameter. Similarly mu attempts to sample its parameters, but just evaluates them since they're constants.

dustinvtran:
That's a great idea. I do think something of this form for the API is very valuable. The challenge is in the implementation.

> + change the shape attribute of RandomVariable to have [..., None] dim representing unknown sample dimension.

[...]

Issue #774, opened 2017-09-30
Extend sample_shape in ed.RandomVariable's to accept tf.Tensor


dustinvtran:
As noted in https://github.com/blei-lab/edward/issues/759#issuecomment-333026192.

PR #778, opened 2017-10-02
Marginal Utility Function

cshenton:
Re: #759 this adds limited support for full graph sampling. Given an RV, `ed.marginal` will traverse its parent graph, replacing any root ancestor instances of `RandomVariable` with a sampled equivalent, so that each non-root RV in the graph is evaluated with a tensor of `batch_shape` of parameters. 

For example:
```python
import edward as ed
import numpy as np

ed.get_session()
loc = Normal(0.0, 100.0)
y = Normal(loc, 0.0001)
conditional_sample = y.sample(50)
marginal_sample = ed.marginal(y, 50)

np.std(conditional_sample.eval())
0.000100221

np.std(marginal_sample.eval())
106.55982
```

This current implementation does not work for graphs of RVs using the sample_shape arg. That will require some refactoring of how `RandomVariable` internally stores the sample_shape. I'm making this PR mostly because I'm confident that the API will be backwards compatible.

Beyond not allowing `sample_shape`, `ed.marginal` can fail in the following ways.

- Failure during graph construction, which could be caused by a successful broadcast in the source graph being broken by the sample dimension.
- Failure after graph construction but before execution, as a result of `ed.marginal` detecting incorrect broadcasting (this prevents situations where sampling `10000` from a scalar RV produces some enthusiastically broadcasted `(10000, 10000, 1)` shaped tensor.

Both of these result in a ValueError.


edward/util/random_variables.py
```diff
@@ -778,3 +778,66 @@ def transform(x, *args, **kwargs):
   new_x = TransformedDistribution(x, bij, *args, **kwargs)
   new_x.support = new_support
   return new_x
+
+
+def marginal(x, n):
```
dustinvtran: Functions should be placed according to alphabetical ordering of function names.

edward/util/random_variables.py
```diff
@@ -778,3 +778,66 @@ def transform(x, *args, **kwargs):
...
+
+  Args:
+    x : RandomVariable.
+      Random variable to perform full graph sample on.
+    n : tf.Tensor or int
+      The size of the full graph sample to take.
+
+  Returns:
+    tf.Tensor.
+    The fully sampled values from x, of shape [n] + x.shape
```
dustinvtran: `x.shape = x.sample_shape + x.batch_shape + x.event_shape`. You replace the sample_shape, so the output should have shape `[n] + x.batch_shape + x.event_shape`. But I guess it currently fails if `x.sample_shape` is non-scalar anyways.

edward/util/random_variables.py
```diff
@@ -778,3 +778,66 @@ def transform(x, *args, **kwargs):
...
+                              "with non scalar sample_shape args.")
+  elif ancestors == []:
+    old_roots = [x]
+  else:
+    old_roots = [rv for rv in ancestors if get_ancestors(rv) == []]
+
+  new_roots = []
+  for rv in old_roots:
+    new_rv = copy(rv)
+    new_rv._sample_shape = tf.TensorShape(n).concatenate(new_rv._sample_shape)
```
dustinvtran: `tf.TensorShape()` fails if `n` is a `tf.Tensor`

tests/util/test_marginal.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import edward as ed
+import numpy as np
+import tensorflow as tf
+
+from edward.models import Normal, InverseGamma
+from tensorflow.contrib.distributions import bijectors
```
dustinvtran: I found this test very intuitive. Great work. One note: you don't use the `bijectors` module

tests/util/test_marginal.py
```diff
@@ -0,0 +1,103 @@
...
+  def test_single_ancestor_batch(self):
+    with self.test_session():
+      loc = Normal(tf.zeros([2, 3, 4]), 1.0)
+      y = Normal(loc, 1.0)
+      sample = ed.marginal(y, 5)
+      self.assertEqual(sample.shape, [5, 2, 3, 4])
+
+  def test_sample_passthrough(self):
+    with self.test_session():
+      loc = Normal(0.0, 100.0)
```
dustinvtran: It's with very low probability this can produce a false negative/positive, but in general you should always set seed in tests when checking randomness.

cshenton:
Creating a Marginal class crossed my mind, in particular when I thought how to implement the API as you described it, first creating the marginal RV and then being able to repeatedly sample it.

My thoughts are that the best way to add that would be to add a marginal method to RandomVariable, which would perform the graph manipulation done here, but attach a placeholder as we discussed in #759. So it could be something like:

```python
rv.sample(10, deep=True)
# raises ValueError: 'must take marginal of rv before deep sampling'
marginal_rv = rv.marginal()
marginal_rv.sample(10)
# works as before, sets placeholder to one then samples 10 from the distribution
marginal_rv.sample(10, deep=True)
# desired functionality, sets placeholder to 10 then samples once from batch result
```

It means more functionality is added to RV, but it makes sense that a function on an RV that returns an RV would be a class method. I'd favour the explicit marginal method since we're doing a copy. 

It also makes me a little less squeamish about all this private attribute access, even though its not on `self`, being on members of the class the method is in will make things easier to fix if breaking changes get made elsewhere on the class.

edward/util/random_variables.py
cshenton: This also came up when I looked into #774, and I think would need to be solved at the same time. Sample shape needs a TensorShape, and there's no nice way to turn tensor `n` into one (I don't think). 

So I think either `sample_shape` needs to be interpreted as 'whatever gets passed to sample' and therefore is stored as a tensor, not a tensorshape. This would solve #774, and I don't think would break much. Another alternative is having `sample_shape` and `sample_shape_tensor` attributes built from the actual tensor representation of the RV.

Let me know if you'd prefer this is implemented in the same PR, I'll push the other changes.

PR #779, opened 2017-10-04
Refactor MCMC to use BayesFlow primitives

dustinvtran:
Following discussion of porting features from Edward <-> BayesFlow.


twiecki:
@dustinvtran Can you elaborate on that decision? I'd also be curious of the differences in scope (and in general) between BayesFlow and Edward.

dustinvtran:
You and me both!

Issue #780, opened 2017-10-04
\epsilon_s


tomisilander:
In http://edwardlib.org/tutorials/klqp in the last equation the \epsilon(s) should be \epsilon_s
ts.

dustinvtran:
Good catch. Contribution welcome.

dustinvtran:
Fixed by #782.

PR #782, opened 2017-10-07
added subscripts to the epsilons in the last equation

gfeldman:
I want to resolve issue #780 so I added the subscripts to the epsilons in the last equation of docs/tex/tutorials/klqp.tex


dustinvtran:
LGTM

Issue #290, opened 2016-10-06
utility function to assess conditional independence


dustinvtran:
for example:

``` python
def is_independent(x, y, z=None):
  """Assess whether x is independent of y given z.

  Implemented using the Bayes Ball algorithm.

  Parameters
  ----------
  x : RandomVariable or list of RandomVariable
    Query node(s).

  y : RandomVariable or list of RandomVariable
    Query node(s).

  z : RandomVariable or list of RandomVariable, optional
    Random variable(s) to condition on.

  Returns
  -------
  bool
    True if x is conditionally independent of y given z.
  """
  raise NotImplementedError()
```

siddharth-agrawal:
@dustinvtran Seems to me that this has been resolved.

dustinvtran:
Another good catch!

PR #791, opened 2017-11-02
Utility function to assess conditional independence

bkayalibay:
This is an attempt at addressing issue #290. The utility function `is_independent` is located at `edward/util/random_variables.py` and I've added some tests at `tests/util/test_is_independent`.


edward/util/random_variables.py
```diff
@@ -714,6 +715,93 @@ def get_variables(x, collection=None):
...
+       Query node(s).
+
+    condition : RandomVariable or list of RandomVariable, optional
+       Random variable(s) to condition on.
+
+  Returns
+    bool
+    True if a is independent of b given the random variables in condition.
+
+  References
```
dustinvtran: Can you add a bibtex entry to `docs/tex/bib.bib`? Also note the bib file is sorted chronologically by year, then sorted alphabetically by author's last name within a year.

edward/util/random_variables.py
```diff
@@ -714,6 +715,93 @@ def get_variables(x, collection=None):
   return list(output)
 
 
+def is_independent(a, b, condition=None):
+  """Assess whether a is independent of b given the random variables in
+  condition.
+
+  Implemented using the Bayes-Ball algorithm[1].
+
+  Args
```
dustinvtran: Fields like `Args` and `Returns` require a colon at the end, c.f., other function docstrings in this file.

edward/util/random_variables.py
```diff
@@ -714,6 +715,93 @@ def get_variables(x, collection=None):
...
+    a : RandomVariable or list of RandomVariable
```
dustinvtran: Args are written as, e.g., `a: RandomVariable or list of RandomVariable.` Note the removed space and added period. When in doubt, you can also double check the generated API docstrings by checking out the README in `docs/`, or consult with the other function docstrings.

bkayalibay:
Thank you very much for the feedback. I've made changes to the docstring and added a bibtex entry. The function actually uses Alg 2 from Schachter (1998). If you can point me towards a more efficient approach I'd be happy to modify the implementation.

dustinvtran:
Perfect! LGTM.

PR #793, opened 2017-11-06
fixed invgamma_normal_mh example

christopherlovell:
Wrong variable in print statement

### DIFF
--- a/examples/invgamma_normal_mh.py
+++ b/examples/invgamma_normal_mh.py
@@ -19,7 +19,7 @@ N = 1000
 loc = 7.0
 scale = 0.7
 xn_data = np.random.normal(loc, scale, N)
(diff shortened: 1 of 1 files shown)


examples/invgamma_normal_mh.py
```diff
@@ -19,7 +19,7 @@
 loc = 7.0
 scale = 0.7
 xn_data = np.random.normal(loc, scale, N)
-print('sigma={}'.format(sigma))
+print('sigma={}'.format(scale))
```
dustinvtran: Can you change `sigma` to scale too?

christopherlovell: changed

dustinvtran:
Thanks!

Issue #211, opened 2016-08-10
Time-series models in edward


fmpr:
Hi,

I'm interested in doing some time-series state-space modelling in edward. So I started implementing a simple autoregressive model AR(k) with a toy dataset. I started out with fixed values for the observation noise and for the variance in the latent process, and everything seemed to work great and fast (code for this example is available here:  [ARk_fixed_noises.py](https://github.com/fmpr/edward/blob/example/timeseries/examples/ARk_fixed_noises.py)  ).

However, if I place Inverse-Gamma priors on these two variances and estimate their distributions, then I need to increase the number of samples significantly in order to obtain any decent results for the inferred states, and then MFVI (naturally) becomes quite slow even for this very small toy example (code with InvGamma priors: [ARk_estimate_obs_noise.py](https://github.com/fmpr/edward/blob/example/timeseries/examples/ARk_estimate_obs_noise.py)  and [ARk_estimate_both.py](https://github.com/fmpr/edward/blob/example/timeseries/examples/ARk_estimate_both.py)  ).

So may questions are: 
- Did anyone implemented time-series models in edward before?
- Do you have any intuition why introducing these two univariate Inverse-Gamma priors can make such a huge difference in the behavior of the inference algorithm? (I have similar models implemented in Stan with ADVI and inference is fast...). Am I doing something wrong? Any suggestions are welcome :-)

akucukelbir:
hi @fmpr,

this is great! we've been meaning to add some time-series models to our examples and website. we would love to work with you on these.

a quick question before we begin to dig into your examples: does ADVI in Stan work well for the Inverse Gamma prior case? (my intuition says that it shouldn't work very well in that case.)

fmpr:
Hi @akucukelbir,

Actually, in my Stan code I had a half-Cauchy prior on the variances, but I just changed that to a Inverse Gamma and it still works great. Once the Stan model is compiled, it converges in just a few seconds (for a dataset that is pretty much the same as the one that I'm trying in edward), and it gives good results. 

Here is the Stan model and the R script that uses it, in case you want to have a look. I was pretty much trying to get something similar to this working in edward as a starting point...
[kf_simple_invgamma.stan](https://github.com/fmpr/edward/blob/example/timeseries/examples/kf_simple_invgamma.stan)
[kf_simple_invgamma.R](https://github.com/fmpr/edward/blob/example/timeseries/examples/kf_simple_invgamma.R)

maedoc:
[...]

PR #802, opened 2017-11-22
[WIP] EM SDE Distribution

maedoc:
This was a comment #211 which snowballed into a PR, which ports [PyMC3](https://github.com/pymc-devs/pymc3/blob/master/pymc3/distributions/timeseries.py#L247) implementation of Euler-Maruyama scheme for stochastic differential equations.  I'll follow up on 

- [ ] batch vs event shaping (I don't really understand this yet)
- [ ] sampling
- [ ] tests/examples

but feedback appreciated


maedoc:
I'm also wondering if I should implement the "non-centered" version of the this, where instead of `N(mu, sd)`, we'd have an auxiliary distribution `N(0,1)` which is then used in the SDE step.  I understand this is better for an HMC sampler, but not sure if the other inference algorithms in Edward would also benefit from that.

dustinvtran:
Apologies for the delay. Was busy at NIPS. 

This seems very useful. If I understand correctly, it's a vectorized approach to handle the time series. The log prob implementation makes sense.

Issue #788, opened 2017-10-29
Confusion in implementation of Metropolis Hastings


hrishikeshv:
In the metropolis_hastings.py file, the docstring for build_update function states that

    \\text{ratio} =
          \log p(x, z^{\\text{new}}) - \log p(x, z^{\\text{old}}) +
          \log g(z^{\\text{new}} \mid z^{\\text{old}}) -
          \log g(z^{\\text{old}} \mid z^{\\text{new}})

However, for MH algorithm we follow the following procedure
![](https://snag.gy/Ks49Xt.jpg)

Hence, shouldn't \log g(z^{\\text{new}} \mid z^{\\text{old}}) be subtracted from the final ratio and \log g(z^{\\text{old}} \mid z^{\\text{new}}) be added. I am attaching for the ratio expression below.

    \\text{ratio} =
          \log p(x, z^{\\text{new}}) - \log p(x, z^{\\text{old}}) -
          \log g(z^{\\text{new}} \mid z^{\\text{old}}) +
          \log g(z^{\\text{old}} \mid z^{\\text{new}})

dustinvtran:
You're correct. I think the implementation is off. IIRC all current examples using MH use a symmetric proposal so it doesn't affect them. Contribution welcome.

PR #806, opened 2017-12-09
Fix the calculation of ratio inside metropolis_hastings.py

siddharth-agrawal:
Minor fix for issue #788. Tested using `tests/inferences/test_metropolishastings.py`.


dustinvtran:
LGTM. Thanks again! Fixes #788.

Issue #807, opened 2017-12-16
Automated transformations break log_joint computation


davmre:
Currently all inference in models with auto-transformed support is silently producing meaningless results. This seems quite serious because automated transformations are enabled by default. 

The problem: when an automated transformation is applied to a latent variable `z`, inference methods that use the `ed.copy(x, dict_swap)` idiom to compute a log joint density (this is basically all inference methods as far as I can tell) will not swap out the transformed variable correctly. Any other variable `x` that depends on `z` will be evaluated using its original dependency---a prior sample from `z`---instead of at a posterior value `qz`. This means that instead of being a deterministic quantity that depends on the posterior, the constructed `log_joint` becomes a random quantity that has no relationship to `qz` outside of the explicit prior term (that term `z.log_prob(qz)` will be computed correctly since it accesses `qz` explicitly rather than through `dict_swap`). This is not caught by tests because none of the tests for automated transformations consider models with multiple variables. 

More specifically, the `Inference` class replaces `z` with `transform(z)` in its `latent_vars` mapping. This is inherited by `dict_swap` which also then contains `transform(z)` rather than `z`. And since `z` isn't in `dict_swap`, its occurences in the graph will never be replaced by `qz`.

I've reproduced this problem with HMC and KLqp inference but it should exist more generally. Here's a test case running HMC on a simple beta-bernoulli model:
```python
import numpy as np
import tensorflow as tf
import edward as ed

# model
z = ed.models.Beta(1., 1., name="z")
xs = ed.models.Bernoulli(probs=z, sample_shape=10)

# data
x_obs = np.asarray([1, 1, 1, 1, 1, 1, 1, 1, 1, 1], dtype=np.int32)

# inference
qz_samples = tf.Variable(tf.random_uniform(shape=(2000,)))
qz = ed.models.Empirical(params=qz_samples, name="z_posterior")
inference_hmc = ed.inferences.HMC({z: qz}, data={xs: x_obs})
inference_hmc.run(step_size=1.0, n_steps=5, auto_transform=True)

z_unconstrained = inference_hmc.transformations[z]
qz_constrained = z_unconstrained.bijector.inverse(qz_samples)
qz_mean, qz_var = sess.run(tf.nn.moments(qz_constrained, 0))

true_posterior = ed.models.Beta(1. + np.sum(x_obs), 1. + np.sum(1-x_obs))
pz_mean, pz_var = sess.run((true_posterior.mean(), true_posterior.variance()))

[...]

PR #808, opened 2017-12-16
Explicitly track constrained and unconstrained versions of transformed variables

davmre:
This is a proposed fix to https://github.com/blei-lab/edward/issues/807 (Automated transformations break log_joint computation). It:

- modifies the `auto_transform` section of `inference.initialize` to maintain two dicts with clear semantics: `latent_vars` maps the original `z`s to `qz`s with matching support (possibly constructed as a transformation of an unconstrained `qz`), while `latent_vars_unconstrained` guarantees to contain versions of the `z`s and `qz`s supported over an unconstrained space. This ensures that any inference method that looks at `latent_vars` to compute a log joint will correctly substitute `z`s with `qzs` in the same space. 
- Updates HMC to wrap the `log_joint` computation by explicitly transforming back into the original (potentially constrained) latent space. I believe HMC needs special attention to work in the unconstrained space because it directly accesses `qz.params`; other inference methods that just treat `qz` as a Distribution (eg variational methods such as KLqp) should automatically do the right thing when given a `qz` having constrained support matching `z`. 
- Adds test cases for a simple beta-bernoulli model with automated transformations using HMC and KLqp.

### DIFF
--- a/edward/inferences/hmc.py
+++ b/edward/inferences/hmc.py
@@ -73,13 +73,25 @@ class HMC(MonteCarlo):
     The updates assume each Empirical random variable is directly
     parameterized by `tf.Variable`s.
     """
-    old_sample = {z: tf.gather(qz.params, tf.maximum(self.t - 1, 0))
-                  for z, qz in six.iteritems(self.latent_vars)}
+
+    # Gather the initial state, transformed to unconstrained space.
+    try:
+      self.latent_vars_unconstrained
+    except:
+      raise ValueError("This implementation of HMC requires that all "
+                       "variables have unconstrained support. Please "
+                       "initialize with auto_transform=True to ensure "
+                       "this. (if your variables already have unconstrained "
+                       "support then doing this is a no-op).")
+    old_sample = {z_unconstrained:
+                  tf.gather(qz_unconstrained.params, tf.maximum(self.t - 1, 0))
+                  for z_unconstrained, qz_unconstrained in
+                  six.iteritems(self.latent_vars_unconstrained)}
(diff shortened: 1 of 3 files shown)


dustinvtran:
Thanks Dave! Merging.

There's an interesting implication of this PR. Namely, ADVI works by transforming prior variables to the unconstrained space, then minimizing trainable Gaussians to the unconstrained posterior. What this PR does is transform trainable Gaussians to the constrained space, then minimize over the original space to the posterior. Parameterization of course matters—though I'd be hard-pressed to point to an example where one approach is preferred over the other (?).

davmre:
Thanks Dustin!

Your point about parametrization is interesting. For HMC this PR still runs
the dynamics in the unconstrained space. For VI, are you sure that
parametrization makes a difference? We're now computing KL[q|p] (or rather,
the ELBO which is equivalent up to a constant) in the original rather than
transformed parametrization, but since KL is an information theoretic
quantity I want to say that this shouldn't change the solution.


On Jan 6, 2018 7:54 PM, "Dustin Tran" <notifications@github.com> wrote:

Merged #808 <https://github.com/blei-lab/edward/pull/808>.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/blei-lab/edward/pull/808#event-1412597449>, or mute the
thread

davmre:
I realized I misstated this slightly, because KL involves a ratio of densities it's invariant to parametrization (https://en.wikipedia.org/wiki/Kullback%E2%80%93Leibler_divergence#Properties) even though other information theoretic quantities like differential entropy aren't.

PR #813, opened 2018-01-01
Add regularization terms in klqp.py

### LINKED ISSUE #529: Add regularization loss in build_loss_and_gradients?
Current implementation of loss in [variational_inference.py](https://github.com/blei-lab/edward/blob/master/edward/inferences/variational_inference.py#L64) does not include  `ops.GraphKeys.LOSSES` and `ops.GraphKeys.REGULARIZATION_LOSSES` in tensorflow graph. So I cannot add regularization loss to the loss used in update. Of course, I can manually add it like `inference.loss += tf.losses.get_total_loss()` after calling `inference.initialize(var_list=tf.trainable_variables())`, but that means I have to modify `inference.train` op as well. Will edward support adding regularization loss in the future release?

siddharth-agrawal:
Added regularization terms in the loss functions and gradient calculations in `klqp.py`. This partially fixes #529. If this change is on the right track, will update the remaining files in the next few commits.

### DIFF
--- a/edward/inferences/klqp.py
+++ b/edward/inferences/klqp.py
@@ -640,14 +640,17 @@ def build_reparam_loss_and_gradients(inference, var_list):
 
   p_log_prob = tf.reduce_mean(p_log_prob)
   q_log_prob = tf.reduce_mean(q_log_prob)
+  reg_penalty = tf.reduce_sum(tf.losses.get_regularization_losses())
 
   if inference.logging:
     tf.summary.scalar("loss/p_log_prob", p_log_prob,
                       collections=[inference._summary_key])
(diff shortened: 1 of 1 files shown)


dustinvtran:
Thanks!

I'm okay with not having this include a unit test. Perhaps if or when you implement it for `ed.MAP`, it would be nice to show they return the same answer. 

Can you add a sentence or two in the `Notes` section of each KLqp class? E.g.,
```
The objective function also adds to itself a summation over all tensors in the `REGULARIZATION_LOSSES` collection.
```

siddharth-agrawal:
@dustinvtran Updated with the relevant changes.

dustinvtran:
Thanks!

PR #817, opened 2018-01-06
Regularization for BiGAN, KLpq, GAN and WGAN inferences

siddharth-agrawal:
Adds regularization in the loss functions for BiGAN, KLpq, GAN and WGAN inferences. This makes further progress towards fixing #529.

### DIFF
--- a/edward/inferences/wgan_inference.py
+++ b/edward/inferences/wgan_inference.py
@@ -31,6 +31,9 @@ class WGANInference(GANInference):
   `discriminator` only to share methods and attributes with
   `GANInference`.
 
+  The objective function also adds to itself a summation over all
+  tensors in the `REGULARIZATION_LOSSES` collection.
+
   #### Examples
 
   ```python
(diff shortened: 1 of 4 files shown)


siddharth-agrawal:
@dustinvtran I'm not confident about the change I made in `klpq.py`. Specifically about the fact that I'm subtracting from the loss instead of adding to it. I did this because the gradients are being taken for `-loss`.

dustinvtran:
What you did is correct for both variational and model parameters in `klpq`.

edward/inferences/wgan_inference.py
```diff
@@ -90,10 +93,14 @@ def build_loss_and_gradients(self, var_list):
                                      list(range(1, gradients.shape.ndims))))
       penalty = self.penalty * tf.reduce_mean(tf.square(slopes - 1.0))
 
+    reg_terms_d = tf.losses.get_regularization_losses(scope="Disc")
+    reg_terms_all = tf.losses.get_regularization_losses()
```
dustinvtran: Any reason this one  and GAnInference uses all but BiGAN's uses `scope="Gen"`?

siddharth-agrawal: I went by how variable lists are created in the respective files. For GAN and WGAN, `scope="Gen"` is not used for creating `var_list` so I assumed it wasn't there.

dustinvtran: That makes sense. Merging now. All looks great.

PR #820, opened 2018-01-07
fix feed_dict update; fixes #811

### LINKED ISSUE #811: Properly Override Feed Dict in Gibbs Sampling
I ran into a problem when attempting a mini-batch sample procedure using `Gibbs` sampling inference.

# What's the Problem

This problem isn't constrained to the below use case, it's simply to prove that the functionality provided by the `ed.Gibbs` may have a small bug in `.update(feed_dict=...)`.

The desired functionality I am looking for is this:

```python
...
x_ph = tf.placeholder(tf.float32, [None, d])
inference = ed.Gibbs(latents, data={x: x_ph})
inference.initialize(n_iter=n_batch * n_epoch)
...
x_batch = next(x_batch)
info_dict = inference.update({x_ph: x_batch}) # should feed the placeholder `x_ph` with `x_batch`
```

And the `stderr` says no:

```stdout
...
  File "test.py", line ..., in main
    x_ph = tf.placeholder(tf.float32, [None, d])
...
InvalidArgumentError (see above for traceback): You must feed a value for placeholder tensor 'Placeholder' with dtype float and shape [?,784]
	 [[Node: Placeholder = Placeholder[dtype=DT_FLOAT, shape=[?,784], _device="/job:localhost/replica:0/task:0/device:CPU:0"]()]]
```

# Probable Cause

According to python `dict` update methods, these [lines](https://github.com/blei-lab/edward/blob/master/edward/inferences/gibbs.py#L98-L101) cause the supplied `feed_dict` argument in `inference.update({x_ph: x_batch})` to be overwritten by the `inference` internal `self.feed_dict`, which may be causing the overwriting of the placeholder?

# Temporary Solution

A work around for me was to update the `inference.feed_dict` before calling the update method.

```python
inference.feed_dict.update({x_ph: x_batch})
info_dict = inference.update()
```

# Possible Permanent Solution

Change this,

```python
# from
# feed_dict.update(self.feed_dict)
# to
self.feed_dict.update(feed_dict)
```

IFF the desired `update` functionality, in the first place, is to allow the user to override conditionals, tf placeholders, etc.

dustinvtran:


wenkesj:
@dustinvtran I noticed the CI build seems to fail on `PEP8-check` only, I believe these don't have to do with #811. It may be that python semantic issues are causing a logical build to go down, which may not be good anyway 🔢

dustinvtran:
Thanks for the spot. Merging this. Fixing PEP8 in separate PR.

PR #823, opened 2018-01-10
Add regularization losses in implicit_klqp, map, wake_sleep

siddharth-agrawal:
This is the final change towards solving #529.


siddharth-agrawal:
@dustinvtran Could you elaborate on the test you wanted for `map.py`? I read the current test and wasn't sure of how to go about it.

dustinvtran:
Check that MAP with a normal-normal model is equivalent to a normal likelihood with Gaussian regularization.

dustinvtran:
Great work! LGTM.

PR #847, opened 2018-02-05
[DOC]: fix typo in the number of iterations in supervised_classification notebook

mirca:
The actual number used in the notebook is `5000` instead of `500`.


dustinvtran:
Thanks!

PR #848, opened 2018-02-05
Fix typo. x.shape => dp.shape

yukinagae:


dustinvtran:
Can you actually rename the rv to be x? Not sure why it's dp.. must have been copypasta bug.

yukinagae:
Sure! I've just renamed the rv to be "x". Maybe copypasta from DirichletProcess ("dp") doc. I also found a similar copypasta in Empirical doc which may be bit confusing, so I fixed it as well.

I will create a new PR if you don't like the branch name "fix-typo-point-mass-doc" because I touched Empirical doc too.

dustinvtran:
This works. Thanks!

PR #849, opened 2018-02-07
Fix progress bar error on windows

### LINKED ISSUE #790: Progress bar encoding error - Windows
Total edge case, but this:
https://github.com/blei-lab/edward/blob/0a995aec7a757812c59246b77452ad0b4f26ce64/edward/util/progbar.py#L83

causes edward to crash on my system:

```
UnicodeEncodeError: 'charmap' codec can't encode character '\u2588' in position 17: character maps to <undefined>
```

Switching to any other character resolves the issue.

abs51295:
Resolves #790


dustinvtran:
Thanks!

PR #870, opened 2018-03-10
Add pyenv's .python-version to .gitignore

patrickeganfoley:
Adds `.python-version` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
+.python-version
 env/
 build/
 develop-eggs/


dustinvtran:
Interesting. Where does this one appear?

patrickeganfoley:
It will show up anywhere you specify a local python version.

```
cd
brew install pyenv
pyenv install -l       # lists available versions of python you can install
pyenv install 3.8-dev
pyenv install 2.7.14
pyenv global 3.8.-dev   #  you are now using 3.8

cd edward
pyenv local 2.7.14        
pyenv version 
```

The `pyenv local` line creates `edward/.python-version` which sets you up to use 2.7.14 in that folder.  

It makes it super easy to manage multiple python installations.

dustinvtran:
Got it. Let's defer on merging it to master. At the moment, `.gitignore` is a mess. In the future, we should probably have it mimick another repo like TensorFlow's.

PR #865, opened 2018-03-08
Implementation of replica exchange MCMC

YoshikawaMasashi:
#861 

Happy to open this pull request. I implemented replica exchange MCMC. I would appreciate it if you would review and comment.

PR #871, opened 2018-03-15
Fix errors in ReplicaExchangeMC

YoshikawaMasashi:
Hi.

I implemented ReplicaExchangeMC in #865. But I found some errors and fixed.

<h3>multiple latent variables</h3>

A code which use multiple latent variables does not work, and this code does not work.
```python
inference = ed.ReplicaExchangeMC({w: q_w, b: q_b}, {w: proposal_w, b:proposal_b},
                                                              data={X: X_train, y: y_train})
inference.run()
```
The cause for this is that callable which return a dict cannot be used in `tf.case`.
The error occurs in this code. In this code `replica_sample[i]` is dict. (Exceptionally, It seem that a dict which has only one variable does work.)
```python
sample_i = tf.case({tf.equal(new_replica_idx[candi], i): _stateful_lambda(
                    replica_sample[i])for i in range(self.n_replica)},
                   default=lambda: replica_sample[0], exclusive=True)
```
Therefore, I fixed this error. I implemented to calculate rates at each replica first. And `tf.case` is not used.

<h3>list for `latent_vars`</h3>

When I use a list for `latent_vars`, an error occurs.
```Python
inference = ed.ReplicaExchangeMC(latent_vars=[x],
                             proposal_vars={x: proposal_x})
inference.run()
```
In `super(ReplicaExchangeMC, self).__init__(latent_vars, data)`, `latent_vars`(list) is converted to `latent_vars`(dict). But before this, `latent_vars`(dict) is needed. I also fixed this.

Thanks

### DIFF
--- a/edward/inferences/replica_exchange_mc.py
+++ b/edward/inferences/replica_exchange_mc.py
@@ -15,7 +15,7 @@ from edward.util import check_latent_vars, copy
 class _stateful_lambda:
   """Class to use instead of lambda.
   lambda is affected by the change of x,
-  so memory_lambdda output x at the time of definition.
+  so _stateful_lambda(x)() output x at the time of definition.
   """
 
   def __init__(self, x):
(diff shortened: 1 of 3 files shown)


edward/inferences/replica_exchange_mc.py
```diff
@@ -54,26 +54,23 @@ def __init__(self, latent_vars, proposal_vars, data=None,
...
+    super(ReplicaExchangeMC, self).__init__(latent_vars, data)
 
     self.n_replica = len(inverse_temperatures)
     if inverse_temperatures[0] != 1:
       raise ValueError("inverse_temperatures[0] must be 1.")
-    self.inverse_temperatures = [tf.convert_to_tensor(inverse_temperature,
-                                 dtype=list(latent_vars.values())[0].dtype)
-                                 for inverse_temperature in
-                                 inverse_temperatures]
+    self.inverse_temperatures = tf.constant(
```
dustinvtran: This fails if `inverse_temperatures` is already a Tensor. Maybe cast?

YoshikawaMasashi: Thank you for the advice. I'll fix.

YoshikawaMasashi:
I added a unit test and refactor the code.
When I perform a unit test that use a list for `latent_vars` of `tf.float64`, an error occurs. So I fixed this bug.

Is it better that I also add unit tests in the same way for other MC (`MetropolisHastings` and so on)?

dustinvtran:
Great! Merging. 

> Is it better that I also add unit tests in the same way for other MC (MetropolisHastings and so on)?

Absolutely.

PR #872, opened 2018-03-18
[add] add unit tests of MC and fix small errors

YoshikawaMasashi:
Hi!

I added not only unit tests of `ReplicaExchangeMC` but also those of other MC.

When I inplemented linear regression test in `ed.Gibbs`, an error occurs (As mentioned in https://discourse.edwardlib.org/t/any-tutorials-on-mcmc/179). I don't commit the test for now, Because It might be the correct behavior.

And, I found that an error occurs when I use a list for `latent_vars` of `ed.Gibbs`.
Also, I found that `ed.Gibbs` and `ed.complete_conditional` can't be used with `tf.float64`. So I fixed these.

Thanks.


dustinvtran:
LGTM. Thanks for the tests (and fix)!

PR #898, opened 2018-05-31
Corrected text on http://edwardlib.org/tutorials/inference-networks

skylogic004:
Previously, the tutorial text did not make the following clear:
(1) what things were the "mean" and what things were the "standard deviation".
(2) what the 10 dimensions were for.
(3) that in Edward's Normal function, loc parameter refers to mu and scale refers to sigma.

### DIFF
--- a/docs/tex/tutorials/inference-networks.tex
+++ b/docs/tex/tutorials/inference-networks.tex
@@ -64,6 +64,8 @@ thus offload part of the computational work to shared precomputation
 and adaptation over time" \citep{stuhlmuller2013learning}.
 
 \subsubsection{Implementation}
+In this example we use an inference network to model MNIST digit classes, 
+approximated by Gaussian distributions for each of the $10$ MNIST digit classes.
 
 Inference networks are easy to build in Edward.
 In the example below, a data point \texttt{x} is a 28 by 28 pixel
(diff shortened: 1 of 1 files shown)


docs/tex/tutorials/inference-networks.tex
```diff
@@ -75,10 +75,10 @@ \subsubsection{Implementation}
 x_ph = tf.placeholder(tf.float32, [M, 28 * 28])
 \end{lstlisting}
 
-We build a neural network using Keras.
-It takes a 28 by 28 pixel image and outputs a $10$-dimensional
-output, one for the mean (\texttt{mu}) and one for the standard
-deviation (\texttt{sigma}).
+We build a neural network using Keras consisting of two dense layers, one for the mean (\texttt{loc}) and one for the standard
+deviation (\texttt{scale}).
+Each dense layer takes a 28 by 28 pixel image and outputs $10$ values, one for each digit class.
```
dustinvtran: replace "Each dense layer" with "The network"

dustinvtran:
Thanks! LGTM up to one revision.

docs/tex/tutorials/inference-networks.tex
skylogic004: Ok. Also, would it be more correct to say "The network...outputs 20 values (10 means and 10 standard deviations)" instead?

Reason: I'm not sure how "network" is defined here. Does the network include the dense layers only or does the network include both the dense layers and the normal?

skylogic004:
Better now?

PR #908, opened 2018-07-16
Update random_variables.py

krishkoushik:
Create tf.Operation by passing in pre-created inputs instead of passing them in later. This is to make it compatible with C_API which is now enabled by default in Tensorflow. Also remove graph._add_op() since this is already done during tf.Operation creation in the most recent version.


dustinvtran:
Excellent. Thanks @krishkoushik!

Will merge after tests pass.

Issue #893, opened 2018-05-10
Import error with tensorflow r1.8 


xiangze:
tensorflow r1.8 does not seem to have "set_shapes_for_outputs" and an error occurs when importing edward.

from tensorflow.python.framework.ops import set_shapes_for_outputs
in edward/edward/util/random_variables.py
https://github.com/blei-lab/edward/blob/master/edward/util/random_variables.py#L15

Just comment out "from tensorflow.python.framework.ops import set_shapes_for_outputs" and following lines does not work.
https://github.com/blei-lab/edward/blob/master/edward/util/random_variables.py#L404

How to be implemented this function without set_shapes_for_outputs?

IanQS:
#891 and #882

larsr:
I replaced the import line ([edward/util/random_variables.py#L15]https://github.com/blei-lab/edward/blob/master/edward/util/random_variables.py#L15)  with 

    from tensorflow.python.framework.ops import set_shape_and_handle_data_for_outputs as set_shapes_for_outputs

as it seems like the function has been renamed. Not sure if it is necessary to use it, but at least this quickfix makes it possible to use tensorflow 1.8 (I haven't run many tests, just got it to work)

devinbostIL:
Why has this not been turned into a pull request?

CMoebus:
I am new to TF and Edward. I conda/pip installed TF and Edward according the "Get Started" recommendations. Then I ran Jupyter and got this same error:

import matplotlib.pyplot as plt
import numpy as np
import tensorflow as tf
hello = tf.constant('Hello, TensorFlow!')
sess = tf.Session()
print(sess.run(hello))
C:\Users\claus\Anaconda3\lib\site-packages\h5py\__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
b'Hello, TensorFlow!'
%matplotlib inline
from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

import edward as ed

from edward.models import Normal

plt.style.use('ggplot')
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-2-da3fcdce0f8a> in <module>()
      4 from __future__ import print_function
      5 
----> 6 import edward as ed
      7 
      8 from edward.models import Normal

[...]

PR #911, opened 2018-07-25
Support recent versions of TensorFlow, one more 

burnpanck:
See issues #882 and #893.
Another pull request (#894) apparently never got merged, maybe because it doesn't fix travis?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,11 @@ matrix:
     include:
         - python: 2.7
         - python: 3.4
+        - python: 3.7
+    env:
+        - TF=1.5.0
+        - TF=1.7.0
+        - TF=1.9.0
 notifications:
     email: false
 before_install:
@@ -37,7 +42,7 @@ install:
   - pip install keras
   - pip install matplotlib seaborn scipy
   - pip install networkx==1.9.1 observations sklearn
-  - pip install tensorflow==1.5.0
+  - pip install tensorflow==$TF
   - pip install pystan
   - pip install nbformat nbconvert jupyter_client jupyter
   - python setup.py install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,14 @@
 sudo: required
 dist: trusty
 language: python
-matrix:
-    include:
-        - python: 2.7
-        - python: 3.4
-        - python: 3.7
-    env:
-        - TF=1.5.0
-        - TF=1.7.0
-        - TF=1.9.0
+python:
+  - 2.7
+  - 3.4
+  - 3.6
+env:
+  - TF=1.5.0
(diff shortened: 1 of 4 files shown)


dustinvtran:
Thanks for working on this! Yes, getting it to work on Travis is very important. It looks like it's still failing? I restarted the build.

burnpanck:
I was able to make it work on TensorFlow 1.7, there was one test failing due to round-off. That one actually contained a bug and was passing just accidentally. I have adjusted the test-case such that it would fail now, but fixed the test. For TF 1.8+ there seems to be a lot more things breaking. I identified a problem in `copy`, where you seem to be constructing the operation first, adding the inputs afterwards. This doesn't work anymore, as TF wants to create the C-implementation in `__init__`. Since I won't have time to fix this now, I'll simply adjust the requirements to include up to TF 1.7, and leave TF 1.8+ for another day/person. As a side-note, Travis is also failing on TF 1.2 despite that being the minimum requirement in `setup.py`. I adjusted these requirements to 1.3.0rc0..1.7.0. However Travis seems to fail on some tests irreproducibly, as the comparison of the last two commits shows (they differ only in the travis.yml and the setup.py, but not the actual code).

diego-ardila-alvarez:
I have been looking at the issue. There are two set of tests failing:

The easy one, in `edward/util/random_variables`, replace: 

``` bij = tfb.Invert(tfb.SoftmaxCentered(event_ndims=1)) -> bij = tfb.Invert(tfb.SoftmaxCentered())```

This because the corresponding signature in `tensorflow.contrib.distributions` changed in the recent versions.

As for the other one, as @burnpanck pointed out, it's an issue with `copy` and recursive graphs (e.g. `tf.scan(lambda a, x: a + x, tf.constant([1.0, 2.0, 3.0]))`). I am trying to find a workaround, as the current logic to handle these cases doesn't seem to work anymore.

diego-ardila-alvarez:
@dustinvtran  Would it make sense to refactor `ed.copy` to use `contrib.graph_editor`? 
https://www.tensorflow.org/api_guides/python/contrib.graph_editor 

I believe the current implementation was possibly based on something like:
 
https://github.com/tensorflow/tensorflow/blob/r1.2/tensorflow/contrib/copy_graph/python/util/copy_elements.py

...Modified to support swapping and recursive graph definitions. However, it also makes `edward` very dependent on the inner workings of `tensorflow`. 

At the moment the implementation is not working with recent `tf` versions, as the workaround for recursive graphs (defining inputs/outputs after the creation of the operation) conflicts with some recent changes. With `contrib.graph_editor` things might become more stable, though online editions are not possible. I have just started to use `edward`, so I don't know if this is too much of a restriction.

PR #924, opened 2018-10-17
Update uses of distributions / bijectors to use TensorFlow Probability.

srvasude:
This pull request updates uses of tf.contrib.distributions with tfp.distributions (TensorFlow Probability: https://www.tensorflow.org/probability/).


srvasude:
It seems like Travis is failing on:

pluggy.manager.PluginValidationError: Plugin 'pytest_cov' could not be loaded: (coverage 4.0.3 (/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages), Requirement.parse('coverage>=4.4'))!

Suggestions?

palikar:
My PR (#922) itself fails at the exact same point. I think the build environment is not set up properly.

